PR #44, opened 2025-06-30
CI: push docker image with perf-maxtext tag

Ruturaj4:
- Added `Dockerfile.maxtext` to extend the existing `jax-ubu22` base image with MaxText source and dependencies.
- Updated GitHub Actions workflow to:
  - Build the MaxText image using the base image (`jax-ubu22.rocm<version>`)
  - Tag the image with:
    - `nightly` for latest builds
    - `<commit_sha>` for traceability
    - `run<github.run_id>` for one-to-one mapping with workflow runs
  - Push the image to GitHub Container Registry (GHCR)
- Adjusted the container launch step to use the correct `run<run_id>` tag.

This enables consistent nightly builds and version tracking for MaxText performance evaluation runs.

### DIFF
--- a/docker/Dockerfile.maxtext
+++ b/docker/Dockerfile.maxtext
@@ -1,0 +1,16 @@
+ARG BASE_IMAGE=jax-ubu22.rocm641
+FROM ${BASE_IMAGE}
+
+# Install git
+RUN apt-get update && apt-get install -y git
+
+# Clone MaxText from rv_jax branch.
+ARG MAXTEXT_BRANCH=rv_jax
+RUN git clone --depth=1 --branch ${MAXTEXT_BRANCH} https://github.com/ROCm/maxtext.git /maxtext
+
+# Set working directory.
+WORKDIR /maxtext
+
+# Install MaxText dependencies.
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.maxtext
```diff
@@ -0,0 +1,16 @@
+ARG BASE_IMAGE=jax-ubu22.rocm641
+FROM ${BASE_IMAGE}
+
+# Install git
+RUN apt-get update && apt-get install -y git
+
+# Clone MaxText from rv_jax branch.
+ARG MAXTEXT_BRANCH=rv_jax
```
charleshofer, line 8: Minor, but I think this would be better to set in the workflow file so that we can just look at the workflow file to tell where maxtext is pulled from. Could we name this just `jax`, instead of having your initials in there, too 😛

charleshofer (approved):
Looks good. Not a huge deal if we fix the branch name before merging. Depends on what we can do with the maxtext repo.

docker/Dockerfile.maxtext
Ruturaj4: ok yes, I need to change the jax branch in the maxtext repo. Maybe change it to `jax_perf`.

PR #64, opened 2025-07-16
Bugfix a bazel debug rule and update gitignore

Arech8:


.gitignore
```diff
@@ -181,3 +181,9 @@ cython_debug/
 
 # PyPI configuration file
 .pypirc
+
+# vscode local settings
+/.vscode
```
Ruturaj4, line 186: what if .vscode or .llvm.sh is not in the root

can we do generic:

```
.vscode
.llvm.sh
```

Ruturaj4: I believe VSCode will respect `.vscode/` inside that folder only when you open that subfolder in VSCode, not the entire repo.

charleshofer (approved):

PR #69, opened 2025-07-17
added three args to docker dev setup script

gulsumgudukbay:
Added the following args:
--rocm_version
--rocm_build_number
--rocm_job_name

### DIFF
--- a/tools/docker_dev_setup.sh
+++ b/tools/docker_dev_setup.sh
@@ -6,8 +6,6 @@ GREEN="\033[32;01m"
 RED="\033[31;01m"
 OFF="\033[0m"
 
-ROCM_VERSION="6.4.0"
-
 info() {
   echo -e " ${GREEN}*${OFF} $*" >&2
 }
@@ -20,6 +18,22 @@ die() {
   [ -n "$1" ] && error "$*"
   exit 1
 }
+
+# Default values
+rocm_version="6.4.0"
+rocm_build_number=""
+rocm_job_name=""
+
+# Parse named command-line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --rocm_version) rocm_version="$2"; shift ;;
+        --rocm_build_number) rocm_build_number="$2"; shift ;;
+        --rocm_job_name) rocm_job_name="$2"; shift ;;
+        *) error "Unknown parameter passed: $1"; exit 1 ;;
+    esac
+    shift
+done
 
 
 install_clang_packages() {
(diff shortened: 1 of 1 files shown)


tools/docker_dev_setup.sh
```diff
@@ -79,9 +93,37 @@ apt-get install -y \
...
+# check if ROCm is installed using rocminfo or fallback to checking /opt/rocm
+if command -v rocminfo &> /dev/null; then
+  info "ROCm is already installed (found via rocminfo). Skipping installation."
+elif [[ -d "/opt/rocm" ]]; then
+  info "ROCm directory found at /opt/rocm. Assuming ROCm is installed. Skipping installation."
+else
+  info "ROCm is not installed. Proceeding with installation..."
+
+  # if major version is 7, then build number and name must be provided
+  if [[ "$major_version" == "7" ]]; then
```
Arech8, line 109: Just like on line `119` this probably should be ` > 6` too.

tools/docker_dev_setup.sh
```diff
@@ -101,13 +143,12 @@ python -m pip install -r \
   build/requirements.txt
 
 # Apply patch for namespace change if ROCm version >= 7
-major_version=$(echo "$ROCM_VERSION" | cut -d. -f1)
 if [ "$major_version" -ge 7 ]; then
```
Arech8, line 146: Perhaps `-gt 6` is better for consistency with the other code reasons.
Also maybe it just makes sense to make a bool var `is_rocm_dev = major_version > 6` and just use the var everywhere instead of comparisons.

Arech8 (approved):
Looks awesome, thanks! Approved!

tools/docker_dev_setup.sh
```diff
@@ -79,9 +93,37 @@ apt-get install -y \
...
+if command -v rocminfo &> /dev/null; then
+  info "ROCm is already installed (found via rocminfo). Skipping installation."
+elif [[ -d "/opt/rocm" ]]; then
+  info "ROCm directory found at /opt/rocm. Assuming ROCm is installed. Skipping installation."
+else
+  info "ROCm is not installed. Proceeding with installation..."
+
+  # if major version is 7, then build number and name must be provided
+  if [[ "$major_version" == "7" ]]; then
+    if [[ -z "$rocm_build_number" || -z "$rocm_job_name" ]]; then
```
Arech8, line 110: Do you think it makes sense to add a default for `rocm_job_name` that we generally use (I think it's `compute-rocm-dkms-no-npi-hipclang`)?

Arech8 (approved):

PR #71, opened 2025-07-21
[master] rocm usecase install doesn't require amdgpu repo addition

pramenku:
docker is already building fine without amdgpu. As such, amdgpu repo should not be framework requirement.

We are adding amdgpu repo with KMD ID and we are not using anywhere installing from this repo. ⁠
So, eventually, amdgpu part is all stale code now.

2+months before KMD is made independent and no more libdrm/mesa packages present at KMD ID repo path due to restructure.

https://osibuilds.amd.com/dashboard/#/release/488/component/3500 path of KMD ID, only amdgpu-dkms and firmware present which we eventually don't install in docker build process.

So, last 2months or so, all these code part is stale and docker is building fine and even tests are running fine.

### DIFF
--- a/build/tools/get_rocm.py
+++ b/build/tools/get_rocm.py
@@ -96,7 +96,6 @@ UBUNTU = System(
 RHEL8 = System(
     pkgbin="dnf",
     rocm_package_list=[
-        "libdrm-amdgpu",
         "rocm-dev",
         "rocm-ml-sdk",
         "miopen-hip ",
@@ -301,6 +300,7 @@ gpgkey=https://repo.radeon.com/rocm/rocm.gpg.key
 """
             % rocm_version_str
         )
+
 
 
 def parse_args():
--- a/build/tools/get_rocm.py
+++ b/build/tools/get_rocm.py
@@ -272,12 +272,6 @@ def setup_repos_ubuntu(rocm_version_str):
     keyadd = "wget -qO - https://repo.radeon.com/rocm/rocm.gpg.key | sudo apt-key add -"
     subprocess.check_call(keyadd, shell=True)
 
-    with open("/etc/apt/sources.list.d/amdgpu.list", "w") as fd:
-        fd.write(
-            ("deb [arch=amd64] " "https://repo.radeon.com/amdgpu/%s/ubuntu %s main\n")
-            % (rocm_version_str, codename)
-        )
-
     with open("/etc/apt/sources.list.d/rocm.list", "w") as fd:
         fd.write(
             ("deb [arch=amd64] " "https://repo.radeon.com/rocm/apt/%s %s main\n")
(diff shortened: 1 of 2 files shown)


pramenku:
With this PR, rocm install worked fine still, https://rocm-ci.amd.com/job/mainline-framework-jax-ub22.04-py3.10/868/console

mrodden:
The script your modifying is meant to be more general than just a Docker install script, so I incorporated your change of function into the main source of truth at https://github.com/mrodden/get-rocm/pull/2/files

Could you take a look and verify? I think due to the --no-dkms option to the amd-installer we were never getting the firmware or driver modules anyway, and I think the only thing we'd get really would have been the libdrm-amdgpu portion, but let me know.

pramenku:
> The script your modifying is meant to be more general than just a Docker install script, so I incorporated your change of function into the main source of truth at https://github.com/mrodden/get-rocm/pull/2/files
> 
> Could you take a look and verify? I think due to the --no-dkms option to the amd-installer we were never getting the firmware or driver modules anyway, and I think the only thing we'd get really would have been the libdrm-amdgpu portion, but let me know.

Thanks @mrodden for review.

Yes, --no-dkms is used so anyway, amdgpu-dkms* packages are not installed.

>> I think the only thing we'd get really would have been the libdrm-amdgpu portion,
This will be pulled from upstream, You don't need to specify it at all. 
"http://rocm-ci.amd.com/job/%s/%s/artifact/amdgpu_kernel_info.txt" -> This has only KMD ID which doesn't contain anymore libdrm* package, which you can see by going at https://osibuilds.amd.com/dashboard/#/job/2193121 ->referring, https://mkmartifactory.amd.com:8443/artifactory/amdgpu-deb-local-new/pool/2193121/jammy/

From ROcm7.0 or even before, we never need mandatory libdrm-amdgpu from amdgpu internal repo.
We can always use upstream which eventually, we are using from 2 months without any issue.

Coming to https://github.com/mrodden/get-rocm/pull/2/files, I am fine if you still want to keep code in function and provide user an option to skip it but keep it default false. Keeping true is concern here. 
FYI even if you are keeping true, it's not at all pulling libdrm-amdgpu from amdgpu internal repo going forward from ROCm7.0 release onwards or mainline, FYI.

We can have your morning 1-1 10mins quick call if needed. I am happy to explain it.

Thanks.

mrodden:
> FYI, Whether your are keeping true/false, it's not at all pulling libdrm-amdgpu from amdgpu internal repo going forward from ROCm7.0 release onwards or mainline, if you use "amdgpu_kernel_info.txt" file.
To really take that effect, you need to use "amdgpu_umd_info.txt" for libdrm.

> amdgpu_kernel_info.txt is stale file now applicable <=6.4.x
> amdgpu_umd_info.txt is applicable >=7.0

Ok... I think for the internal installs I will probably just remove the block that would install the driver and the libdrm-amgpu packages then.

PR #79, opened 2025-07-24
adding update to get_rocm.py script to update the apt packages

gulsumgudukbay:
Before rocm is installed, apt install wget was throwing error. So before install, we need to run apt update.


charleshofer:
This fixes what I broke in this PR: https://github.com/ROCm/rocm-jax/pull/60/files#diff-1a854cba1feb37e61e5438a5f1246681a7d4e3f96063d3fa2b41e1b3c42138bcL77. Not sure why CI didn't complain about it. Related to caching maybe?

charleshofer (approved):

PR #86, opened 2025-08-04
Clean up week-old docker things

charleshofer:
Clean up docker build cache, images, stopped containers, and networks older than a week. We have limited space on our CI machines, and once it's filled, CI jobs start to fail. Usually, this is just old docker artifacts that have to be deleted with `docker system prune`. This runs a prune before we build anything with docker.


psanal35 (approved):
Looks good to me~

PR #87, opened 2025-08-05
Tell the docker build script to tag builds as 'nightly'

charleshofer:
Docker images weren't getting tagged with the `nightly` tag. We were just creating the `extra-cr-tag` variable in the matrix. We need to actually pass it into the re-usable workflow.


psanal35 (approved):
LGTM~

PR #92, opened 2025-08-12
bump up xla according to xla team's request

Ruturaj4:
bump up xla according to xla team's request

@charleshofer Some improvements:

TF_XLA_HSACO_BITCODE_SIZE_THRESHOLD defines bit code size threshold (set to 64K by default) deciding if a particular LLVM module shall be cached
TF_XLA_HSACO_CACHE_DIR defines a hsaco cache location (defaults to /tmp).
use flat_hash_map for in-memory hsaco cache to prevent adding duplucate entries
use ofstream read/write instead of RenameFile if hsaco cache file cannot be moved
numerous code cleanups


i-chaochen:
please not to add my name on PR title...

i-chaochen (approved):

charleshofer:
You don't need to tag me in the description. You can just write a description. But this is way better. Thanks.

PR #95, opened 2025-08-19
Add commit info to builds

AratiGanesh:
Motivation: Adding commit hashes (ROCm, XLA, JAX) to the build scripts, wheels and Docker files will make triage easier.
Changes: 
- Updated ci_build to accept and pass commit metadata
- Modified Dockerfiles to set these values as labels
- Updated wheel building process to include commit metadata in build artifacts

### DIFF
--- a/jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
+++ b/jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
@@ -24,9 +24,12 @@ import pathlib
 import stat
 import subprocess
 import tempfile
+import textwrap
+import json
 
 from bazel_tools.tools.python.runfiles import runfiles
 from pjrt.tools import build_utils
+
 
 parser = argparse.ArgumentParser()
 parser.add_argument(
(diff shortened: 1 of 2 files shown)


build/ci_build
```diff
@@ -53,29 +61,119 @@ def dist_wheels(
...
+        commit_date_raw = subprocess.check_output(date_cmd, cwd=repo_path).decode("utf8").strip()
+        commit_date = commit_date_raw.split()[0] 
+        
+        return {
+            "commit": commit_hash,
+            "commit_title": commit_title,
+            "commit_date": commit_date
+        }
+        
+    except subprocess.CalledProcessError as ex:
```
Arech8, line 98: Would be nice to issue some diagnostic message at least, if that happens

build/ci_build
```diff
@@ -53,29 +61,119 @@ def dist_wheels(
...
 
-    if xla_path:
-        try:
-            xla_commit = subprocess.check_output(cmd, cwd=xla_path)
-        except Exception as ex:
-            LOG.warning("Exception while retrieving xla_commit: %s" % ex)
+    for repo_id, cfg in repos.items():
+        p = cfg["path"]
+        info = get_repo_commit_info(str(p), repo_id) if p else {
+            "commit": "unknown", "commit_title": "unknown", "commit_date": "unknown"
```
Arech8, line 127: There are multiple uses of "unknown"  dict `{"commit": "unknown", "commit_title": "unknown", "commit_date": "unknown"}`. What if at some point you'd need to add an element to it?

Perhaps you can assign it to a "constant" and just use the constant where these "unknown"s are needed?

build/ci_build
```diff
@@ -140,10 +253,72 @@ def build_dockers(
...
+        if plugin_title != 'unknown':
+            print(f"rocm/rocm-jax Commit Title: {plugin_title}")
+            print()
+
+        
+        print("=" * 70)
+        print()
+        
+    except subprocess.CalledProcessError:
+        print(f"Failed to inspect Docker image: {image_name}")
```
Arech8, line 310: diagnostics message is nice, but how a caller would understand that something wrong has happened, if you neither returning error code, nor translating/re-raising the exception?

I think the simplest acceptable way to deal with it here is to just re-raise the exception. This will eventually crash the python process, which is hard to miss.

build/ci_build
```diff
@@ -21,6 +21,14 @@ import os
...
 import sys
+import datetime
+import json
+from pathlib import Path
+
+
+
+def repo_root() -> Path:
+    return Path(__file__).resolve().parent.parent
 import tempfile
```
Arech8, line 32: generally, imports go first, and then go function definitions...

jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
Arech8, line 1: Changes to this file are suspiciously similar to changes made to the other `build_gpu..py` file. Maintaining this code would be quite unpleasant as identical changes has to be made to 2 files instead of 1. Perhaps you can refactor code that deals with build metadata to a single file and just use it whenever it's needed?

build/ci_build
Arech8: Yeah, agree on that. I missed that's a part of CI build, thanks

jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
Arech8: Yeah, these duplication was a bad idea from the start, but at least we shouldn't make it worse and aim to fix it instead. The suggested location seems for me totally suitable for the task.

build/ci_build
```diff
@@ -53,29 +61,119 @@ def dist_wheels(
...
+
+    try:
+        if not repo_path or not os.path.exists(repo_path):
+            return {
+                "commit": "unknown",
+                "commit_title": "unknown", 
+                "commit_date": "unknown"
+            }
+        
+        if not os.path.exists(os.path.join(repo_path, ".git")):
```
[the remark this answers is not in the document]
AratiGanesh: Noted, I will get rid of the use case when .git does not exist.

PR #100, opened 2025-08-27
Remove verbose logging from auditwheel

charleshofer:
We don't need `auditwheel` to log as much as it does. It's taking up almost half of the log and is mostly output that we thought would be useful for debugging problems with the wheels but hasn't in fact been that useful.


psanal35 (approved):
Looks good to me~

PR #102, opened 2025-08-28
Add pip install to Ubuntu 22 docker build

charleshofer:
Installs pip3 when we install Python. I'm not sure how this docker file was able to build before.

Does some other minor fixes to make sure that we don't get path problems when we have weirdly named directories.


Arech8 (approved):

Ruturaj4 (approved):
LGTM

PR #104, opened 2025-09-04
Bake commit info wheels

AratiGanesh:
### Main Changes - 
- Get the commit hashes directly through workspace.bzl files or from local .git directories through rev-parse
- Output commit_info.py file for the wheels
- Update the Dockerfiles with new args and labels 

### Build Method 1 - Local Development Build

In this case, both the local xla or both local jax and xla are used. 

For local development, use `stack.py`:
```bash
python3 stack.py develop --rebuild-makefile

make clean dist install
```
**Wheels will be generated in /jax_rocm_plugin/dist. Unzip wheels and inspect commit_info.py to verify commit hashes.**

### Build Method 2 -  CI Build
1. python3 build/ci_build     --python-version 3.12     --rocm-version    dist_wheels 
Unzip wheels in /jax_rocm_plugin/wheelhouse and inspect the commit_info.py. The hashes should correspond to the pinned hashes. 

2. python3 build/ci_build     --python-version 3.12     --rocm-version     --xla-source-dir=/path-to-xla dist_wheel
**Unzip wheels in /jax_rocm_plugin/wheelhouse and inspect the commit_info.py. The XLA hash would correspond to the latest commit in the local XLA repo** 

3.  python3 build/ci_build  --python-version 3.12  --rocm-version   build_dockers
**Inspect the generated docker image and verify the commit hash labels.**  [docker inspect <image_name>]

### How to query?
```
python -c "from jax_plugins.xla_rocm7 import commit_info; print(f'
XLA: {commit_info.xla_commit}, JAX: {commit_info.jax_commit}, ROCm-JAX: {commit_info.roc
m_jax_commit}')"
```
XLA: 104647d7abd13d81b9050184ababa3154563eef5, JAX: 5712de44e97c455faed1fd45532e821ca66d025a, ROCm-JAX: c99c033768ef096fd3124f77de0819263f56f841

### DIFF
--- a/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
+++ b/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
@@ -27,7 +27,7 @@ import subprocess
 import tempfile
 
 
-# pylint: disable=import-error,invalid-name
+# pylint: disable=import-error,invalid-name,consider-using-with
 from bazel_tools.tools.python.runfiles import runfiles
 from jaxlib_ext.tools import build_utils
 
--- a/jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
+++ b/jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
@@ -26,7 +26,7 @@ import stat
 import subprocess
 import tempfile
 
-# pylint: disable=import-error,invalid-name
+# pylint: disable=import-error,invalid-name,consider-using-with
 from bazel_tools.tools.python.runfiles import runfiles
 from pjrt.tools import build_utils
 


jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
```diff
@@ -98,8 +104,8 @@ def prepare_wheel_rocm(
   )
   build_utils.update_setup_with_rocm_version(sources_path, rocm_version)
   write_setup_cfg(sources_path, cpu)
-
   plugin_dir = sources_path / f"jax_rocm{rocm_version}_plugin"
+  build_utils.write_commit_info(plugin_dir, args.xla_commit, args.jax_commit, args.rocm_jax_git_hash)
```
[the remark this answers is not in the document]
Arech8: Note the same applies to the other script

jax_rocm_plugin/jaxlib_ext/tools/BUILD.bazel
```diff
@@ -17,9 +17,14 @@ licenses(["notice"])  # Apache 2
 
 package(default_visibility = ["//visibility:public"])
 
+load("//third_party/xla:workspace.bzl", xla_commit = "XLA_COMMIT")
```
Arech8, line 20: How would this work for an overridden XLA, when everything is compiled with `--override_repository=xla=/my/custom/path` ?

cc @charleshofer

Arech8:
Also please always format the code you're proposing to merge and make CI green.
To format Python code in VSCode you can install `Black Formatter` extension from `Microsoft`, or `Ruff` from `Astral Software` ([Ruff](https://docs.astral.sh/ruff/) is fully compatible with `black`, but faster), and then use `Format Document` feature.

build/ci_build
```diff
@@ -22,6 +22,34 @@ from pathlib import Path
...
+
+def _get_commit_info_from_wheel():
+    wheel_file = glob.glob("wheelhouse/*jax_rocm7_plugin*.whl")[0]
+
+    with zipfile.ZipFile(wheel_file, "r") as wheel:
+        commit_content = wheel.read("jax_rocm7_plugin/commit_info.py").decode("utf-8")
+
+    namespace = {}
+    exec(commit_content, namespace)
+    print(namespace["xla_commit"])
```
charleshofer, line 46: nit: Can we somehow indicate to the user what we're printing out? Like,

```
print("This wheel built with XLA commit: ", namespace("xla_commit"])
```

docker/Dockerfile.jax-ubu22
```diff
@@ -58,7 +58,7 @@ RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
         matplotlib \
         absl-py \
         flatbuffers \
-        hypothesis
+        hypothesis
```
charleshofer, line 61: You gotta remove that trailing space. Pylint will likely complain about it.

jax_rocm_plugin/jaxlib_ext/tools/BUILD.bazel
charleshofer: It won't. This only works for the pinned XLA, and we probably should be able to accommodate it. The annoying thing is that I don't see a solution that will work for both pinned XLA and overridden XLA. This just gets whatever is in the XLA_COMMIT variable in the workspace.bzl file. To accommodate override XLA, inside of the build_gpu_kernels_wheel.py, we could try and look for a `.git/ORIG_HEAD` inside of the XLA directory. Unfortunately, the archives that we download with pinned XLA don't contain a .git directory (I think that'll change with Zahid's PR that uses a different bazel rule to download XLA though).

Arech8: Yeah... Not just "only for the pinned", but what's way worse - if XLA repo is overridden, this will cause a totally wrong and misleading commit info used.
If a client would use a custom compiled version, and find and report a bug in it, we'll be in a deep trouble working with a wrong commit. We need to detect a repo override at least...
I know a very hacky way to do that - by checking if `$(bazel info output_base)/external/xla` is a real directory (then it should be a checkout from the pinned commit and commit data is correct, the dir itself *should not* have `.git` subdirectory), or it's a symlink (then it's an overridden directory and ***if*** it has `.git` subdirectory, - use it to query the commit, otherwise make the commit info "unknown"). But not 100% sure it's super robust (though it should be as that's how bazel is documented to work) and it doesn't look to me that implementing this logic is trivially simple inside a bazel itself... But anyway, I think we must as least make sure that if a commit info is present, it is correct. There's nothing worse than a misleading information.

charleshofer: Yeah, this should be addressed in PR. None of those solutions sound super robust though. The best way would probably be to check if `--override_repo=xla=...` was set. If so, we can try and look for a `.git` file in the passed XLA directory. If not, we can set the git commit to something like `"custom"` or maybe `None`. I'd bet that Bazel has a way to look at the options that are set from the command line, but I'm not sure how to do that.

i-chaochen: Then, I think we could add this into a jax function like you desecribed https://github.com/ROCm/rocm-jax/pull/104#issuecomment-3290893872 we could just query this `jax.xla_commit()` to see whether is overridden or its commit info?

jax_rocm_plugin/build/build.py
```diff
@@ -459,12 +459,17 @@ async def main():
     }
     target_cpu = wheel_cpus[args.target_cpu] if args.target_cpu is not None else arch
 
+    logging.error("Local XLA path: %s", args.local_xla_path)
```
Arech8, line 462: Why this is logged as `error` instead of an `info` or smth like that?

AratiGanesh: ah forgot to that that off! Fixed it

build/ci_build
```diff
@@ -22,6 +22,37 @@ from pathlib import Path
...
+def _get_labels(image_name):
+    out = subprocess.check_output(
+        ["docker", "inspect", "--format", "{{json .Config.Labels}}", image_name]
+    )
+    labels = json.loads(out)
+    return labels
+
+
+def _get_commit_info_from_wheel():
+    wheel_file = glob.glob("wheelhouse/*jax_rocm7_plugin*.whl")[0]
```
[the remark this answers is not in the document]
AratiGanesh: Fixed it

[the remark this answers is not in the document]
Arech8: Arati... `glob.glob("wheelhouse/*jax_rocm7_plugin*.whl")` still finds only namespace v7. Gabe's suggestion had one additional star in it in place of version. I don't think globbing stars in front of and beyond are useful, but the version variability is the only reason for the thread the first place...

Arech8:
@AratiGanesh can you please test and update testing info in the PR description how commit baking work with development workflow based on `stack.py` (https://github.com/ROCm/rocm-jax/blob/master/DEVSETUP.md) ?

jax_rocm_plugin/jaxlib_ext/tools/BUILD.bazel
charleshofer: The "custom" XLA case is handled in `build_gpu_kernels_wheel.py` at line 111, not in this file.

charleshofer (approved):
After you address @Arech8's comments, it's fine with me if you merge. Make sure to do a "Squash and merge" when you do since there's more than one commit on this PR.

jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
```diff
@@ -62,112 +64,166 @@
...
 
 [bdist_wheel]
 plat_name={tag}
-python-tag=py3
 """
-    )
+        )
+
+
+def get_xla_commit_hash():
```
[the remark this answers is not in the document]
Arech8: Hmm.... `./jax_rocm_plugin` already have `./build` sub-directory for things related to building. Why not put this file there?

Stepping back, let me clarify. The context of my suggestion was this: "Prior to your PR there already was a code duplication that make maintenance complicated. Planned changes due to the PR added more code duplication hence complicating the maintenance even more". So, I asked you to refactor at least you own new code to at least not increase the amount of duplicated code with the PR. Given that your new code is specifically related to baking or extracting commit info, this looks to me like a great candidate for a single individual component, responsible for doing that task only. If you could do this, that'll great already.

While technically this out of the scope of the PR and I didn't ask for it, but if you can even reduce the amount of prior code duplication, that would be just awesome. But this is a bigger scope task that might require even more work that is seen on the surface now. Perhaps it's wise to merge only commit info baking first, and address that task later. But as I said - it's up to you. My primary concern was increased code duplication. Reducing it is always nice, but is optional due to a bigger effort required.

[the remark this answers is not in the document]
AratiGanesh: @Arech8 appreciate your comment. I restructured the code a bit, although it might not solve the code duplication problem. I will create a different PR to probably move it into the build directory.

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
```diff
@@ -58,120 +60,179 @@
...
+                with open(ref_file_path, "r", encoding="utf-8") as ref_f:
+                    xla_commit_hash = ref_f.read().strip()
+            else:
+                xla_commit_hash = head_content
+
+            print(f"Using local XLA commit hash: {xla_commit_hash}")
+
+        except (FileNotFoundError, IOError):
+            xla_commit_hash = "Custom"
+            print("Could not read XLA commit hash, using 'Custom'")
```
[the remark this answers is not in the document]
AratiGanesh: Have addressed this in the latest commit.

build/ci_build
```diff
@@ -148,6 +187,16 @@ def build_dockers(
...
 
+    labels = _get_labels(image_name)
+
+    print("ROCm JAX Build Information")
+    if "com.amdgpu.rocm_jax_commit" in labels:
+        print(f"rocm/rocm-jax commit hash: {labels['com.amdgpu.rocm_jax_commit']}")
+    if "com.amdgpu.jax_commit" in labels:
+        print(f"rocm/jax commit hash: {labels['com.amdgpu.jax_commit']}")
+    if "com.amdgpu.xla_commit" in labels:
+        print(f"rocm/xla commit hash: {labels['com.amdgpu.xla_commit']}")
```
Arech8, line 198: Imagine you're seeing a string

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
```diff
@@ -58,120 +60,176 @@
...
+"""
+        )
+
+
+def get_xla_commit_hash():
+    """Determines the XLA commit hash to use - local repository or a pinned."""
+
+    # XLA is mounted at /xla if built using ci_build else at /rocm-jax/xla if built using stack.py
+    if args.use_local_xla:
+        return build_utils.get_local_git_commit(["/xla", "/rocm-jax/xla"])
```
Arech8, line 120: There's a problem here, this isn't true. XLA people specifically asked to implement a `stack.py` feature to allow use of any directory as XLA home, and this was [implemented](https://github.com/ROCm/rocm-jax/blob/0e7aa2e3861055d7d5be4c4855dc99bda6281d8b/stack.py#L416). The same applies to JAX. One just can't hardcode these paths here and assume it'll work.

Instead of using a boolean flag `use_local_xla` why just not make use a path argument? You can get a value for such a flag in the same way you're getting a value for `use_local_xla` - by inspecting `--override_repository=xla=` parameter of `--bazel_options`. The easiest way to solve this is to assume the directory never have spaces and quotation marks: for that `--override_repository=xla=([^$\s]+)` is enough (match anything except an end of a string or a space at least once).

I made a version of a regexp works for spaces in the path and supports quotation marks, but due to a dependence on possessive quantifiers needed for the correctness, it requires at least py3.11 (which we can't assume for the build system, if I'm not mistaken), and the regexp itself could spontaneously summon demons or provoke a heart attack in people less used to regexps, so we could probably go with a simple version above (just put a comment that it doesn't support spaces in path, and verify the dir you've obtained exists with an assertion, after you convert it to an abs path with `if not os.path.isabs(path): os.path.realpath(os.path.expanduser(path))`

jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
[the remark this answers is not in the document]
Arech8: ok, lets postpone it until the other PR...Marking this thread as resolved...

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
Arech8: Also please remember you can't assume this script is always called with a correct parameters, so abs path conversion + dirs existence check is necessary in this (and the dupe) file too.

AratiGanesh: @Arech8 I wasn't aware of the --xla-dir addition. So it was assumed the path was fixed in the Makefile. In that case, yes I will read the whole directory from the bazel arguments.

PR #118, opened 2025-09-18
update XLA commit hash to the recent commit on 9/17/2025

gulsumgudukbay:
This PR updates the XLA commit hash to 3b077fcf723b3dda24868f603419d0c914f478e0, which was pushed on 9/17/2025. 
The commit hash was tested and recent wheels are built using this.


gulsumgudukbay:
> Would be good if you could include a more substantial description of the PR, but not going to hold up review for it

updated

PR #125, opened 2025-09-22
Store JAX commit SHA in DB in nightly runs

psanal35:
Since the tests are pulled from JAX repository, this PR also stores the JAX commit hash in the database to better track results against changes.


psanal35:
This feature already in the current DB.

PR #127, opened 2025-09-25
Allow user-provided ROCm and calculate namespace correctly in more places

gabeweisz:
There were some places where the ROCm path and namespace (60 or 7) were still wrong. This fixes them.


jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -172,28 +183,6 @@ def build_jaxlib_wheel(
     _run_scan_for_output(cmd, pattern, env=env, cwd=jax_path, capture="stderr")
 
 
-def build_jax_wheel(jax_path, python_version):
```
gabeweisz, line 175: dead code

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -42,6 +43,19 @@
 )
 
 
+# pylint: disable=R0801
```
gabeweisz, line 46: This is not ideal - there are 3 versions of this code now. I tried to put it in the existing jax_plugin_rocm/build/tools/utils.py but could not get this instance to be able to see it from pylint. I'm open to other suggestions for where to put it.

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -114,9 +128,15 @@ def find_clang_path():
     return None
 
 
-# pylint: disable=R0913, R0917
+# pylint: disable=R0913, R0917, too-many-locals
```
gabeweisz, line 131: Fixing the too-many-locals lint error required inlining compiler == "clang" in the two places where it was used, which to me is less readable than this

PR #134, opened 2025-09-30
Handle tests/ prefix in pytest nodeid parsing

psanal35:
Pytest can return different nodeid formats, and with skip.ini file it is observed that:

- with `-c pytest_skips.ini`: tests/<test_name>.py
- without skip.ini: <test_name>.py

This PR updates parse_test_log() to add "tests/" prefix only when missing.


gulsumgudukbay (approved):
Looks good to me, @psanal35  tested this using CI.

PR #135, opened 2025-10-01
Adding rpath for /opt/rocm/libs since librccl.so etc is not found

gulsumgudukbay:
## Motivation

When testing the built 0.6.0 wheels for DLM e2e workloads, the jax_nlp_seq gives the following error:
```
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: Failed to open /opt/venv/lib/python3.12/site-packages/jax_plugins/xla_rocm7/xla_rocm_plugin.so: librccl.so.1: cannot open shared object file: No such file or directory
E0926 14:22:41.853117 140424214503552 xla_bridge.py:566] Jax plugin configuration error: Exception when calling jax_plugins.xla_rocm7.initialize()
```

## Technical Details

The reason for this error is that the wheel rpath is missing /opt/rocm/lib path which contains the .so files. Therefore, the library is not found. 
To fix this, this PR adds the /opt/rocm/lib to the wheel rpath. 

## Test Plan

This was tested using DLM jax_nlp_seq and other workloads as well.

## Test Result

The newly built wheels do not have this error. 

## Submission Checklist

### DIFF
--- a/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
+++ b/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
@@ -17,6 +17,12 @@
 
 # Most users should not run this script directly; use build.py instead.
 
+# pylint: disable=R0801
+
+"""
+Script to build a JAX ROCm kernel plugin wheel. Intended for use via Bazel.
+"""
+
 import argparse
 import functools
 import os
@@ -25,9 +31,11 @@ import stat
 import subprocess
 import tempfile
 
+# pylint: disable=import-error
 from bazel_tools.tools.python.runfiles import runfiles
 from jaxlib_ext.tools import build_utils
 
+# pylint: enable=R0801
 parser = argparse.ArgumentParser()
 parser.add_argument(
     "--output_path",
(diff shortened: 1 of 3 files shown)


JehandadKhan (approved):

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
```diff
@@ -142,7 +142,7 @@ def prepare_wheel_rocm(
       f"_triton.{pyext}",
       f"rocm_plugin_extension.{pyext}",
   ]
-  runpath = '$ORIGIN/../rocm/lib:$ORIGIN/../../rocm/lib'
+  runpath = '$ORIGIN/../rocm/lib:$ORIGIN/../../rocm/lib:/opt/rocm/lib'
```
i-chaochen, line 145: do we really need to hardcoding the rocm path as `/opt/rocm` ?

jax_rocm_plugin/pjrt/tools/build_gpu_plugin_wheel.py
```diff
@@ -129,7 +129,7 @@ def prepare_rocm_plugin_wheel(sources_path: pathlib.Path, *, cpu, rocm_version):
     raise Exception(mesg) from ex
 
   shared_obj_path = os.path.join(plugin_dir, "xla_rocm_plugin.so")
-  runpath = '$ORIGIN/../rocm/lib:$ORIGIN/../../rocm/lib'
+  runpath = '$ORIGIN/../rocm/lib:$ORIGIN/../../rocm/lib:/opt/rocm/lib'
```
i-chaochen, line 132: ditto

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
charleshofer, line 145: I know that in the past we've assumed that ROCm will always be installed under `/opt/rocm/`, but that isn't going to always be the case going forward. TheRock and ROCm wheels will be installed in other locations, and our wheels are going to have to handle that. Not sure if you can do it with rpath, but can you check the `ROCM_HOME` environment variable for the ROCm path?

charleshofer (changes requested):

jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
gulsumgudukbay: > I know that in the past we've assumed that ROCm will always be installed under `/opt/rocm/`, but that isn't going to always be the case going forward. TheRock and ROCm wheels will be installed in other locations, and our wheels are going to have to handle that. Not sure if you can do it with rpath, but can you check the `ROCM_HOME` environment variable for the ROCm path?

I removed --force-rpath for enabling users to be able to set LD_LIBRARY_PATH

charleshofer (approved):

PR #140, opened 2025-10-08
Init Llama performance test

psanal35:
This PR sets up the llama-perf workflow for Llama performance testing.


charleshofer (approved):
Just add a description and you're good to merge. This is just a stub so that we can test out the real workflow anyways.

PR #143, opened 2025-10-14
The ROCm plugin should fail gracefully 

JehandadKhan:
## Motivation

A customer installs both the CUDA PJRT/Plugin wheels and the ROCM PJRT/Plugin wheels in the same env. However, the current behavior of the ROCm wheels is that they cause a crash which means that the CUDA GPUs or CPU cannot be used, requiring a special JAX build for ROCm. 

## Technical Details

The current infra throws in C++ PJRT which cannot be caught gracefully in the current system, however, if we check if GPUs are available and then throw in python, then JAX handles the situation more gracefully. The same behavior is adopted by CUDA 

## Test Plan

Manually tested 

## Test Result

Pass

## Submission Checklist

- [X ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/jax_rocm_plugin/pjrt/python/__init__.py
+++ b/jax_rocm_plugin/pjrt/python/__init__.py
@@ -136,6 +136,10 @@ def initialize():
 
     set_rocm_paths(path)
 
+    if rocm_plugin_extension is None:
+        logger.warning("rocm_plugin_extension not found")
+        return
+
     device_count = rocm_plugin_extension.get_device_count()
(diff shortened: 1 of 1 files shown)


gulsumgudukbay (changes requested):

gulsumgudukbay (approved):
Looks good. The other comments Copilot generated are not that important in terms of affecting functionality.

PR #146, opened 2025-10-16
Remove the remaining namespace patches as they are no longer needed

psanal35:
There are some updates made by PR https://github.com/ROCm/rocm-jax/pull/144, including removal of namespace patch that are no longer needed. However, a few remained and resulted in an issue in a workflow run: https://github.com/ROCm/rocm-jax/actions/runs/18549183529. This PR handles them.


mminutoli:
> I tested it --it works. I plan to remove the namespace.patch file too.

Can we just do it?

mminutoli (approved):
LGTM. Let's wait for @gulsumgudukbay to check it before merge.

gulsumgudukbay (approved):
Looks good. A note so that we don't get confused in the future: This PR removes the namespace patch that is applied to jaxlib. Since in the requirements file now we are installing jaxlib from our release artifacts - that already has the patch- we do not need patching in dockerfiles.

PR #149, opened 2025-10-18
Update Llama performance workflow by adding a new runner, a model, and JAX version support 

psanal35:
This PR have updates to the existing Llama performance benchmarking workflow (https://github.com/ROCm/rocm-jax/pull/141). It now runs with two different JAX versions:
- 0.6.0: uses official released wheels and Docker image.
- 0.7.1: uses wheels and Docker image built from the latest nightly results.

A follow-up PR will be opened to address the comments below.

### DIFF
--- a/.github/workflows/llama-perf.yml
+++ b/.github/workflows/llama-perf.yml
@@ -227,8 +227,8 @@ jobs:
       - name: Run model
         run: |
           docker exec model-run-jax${{ matrix.jax-version }}-${{ matrix.model-name }} bash -lc '
-            wget -O /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/\
-              releases/download/v1.26.0/bazelisk-linux-amd64
+            BIN_URL="https://github.com/bazelbuild/bazelisk/releases/download/"
+            wget -O /usr/local/bin/bazel "$BIN_URL/v1.26.0/bazelisk-linux-amd64"
             chmod +x /usr/local/bin/bazel
             bazel version
             apt update && apt install -y libdw1 libglib2.0-0


.github/workflows/llama-perf.yml
```diff
@@ -1,140 +1,276 @@
...
+            -e NVTE_USE_ROCM=${{ env.NVTE_USE_ROCM }} \
+            -e HIP_PLATFORM=${{ env.HIP_PLATFORM }} \
+            -e NVTE_ROCM_ARCH=${{ env.ROCM_ARCH }} \
+            -e NVTE_FUSED_ATTN_AOTRITON=${{ env.NVTE_FUSED_ATTN_AOTRITON }} \
+            -e CU_NUM=${{ env.CU_NUM }} \
+            -e PIP_BREAK_SYSTEM_PACKAGES=1 \
+            -v "${{ github.workspace }}:/workspace" \
+            ${{ matrix.docker-image }} \
+            tail -f /dev/null
+      - name: Build TE-wheel
```
charleshofer: The Run Container ..., Build TE-wheel, and Cleanup containers steps can all be combined into one step that uses `docker run --rm`. I'd very much prefer that since it's much easier to read and you don't have to worry about messing up the cleanup step and leaving a bunch of junk around on the runner's system. I know that the other perf workflow does it like that, and I'd wish we'd changed it before it was merged.

.github/workflows/llama-perf.yml
```diff
@@ -1,140 +1,276 @@
...
         run: |
-          ls -la ${{ github.workspace }}/llama-perf/wheelhouse
+          ls -la ${{ github.workspace }}/wheelhouse
       - name: Disable non-root check
         run: |
-          sed -i '/python.toolchain(/a\    ignore_root_user_error = True,' \
-          "${{ github.workspace }}/llama-perf/MODULE.bazel"
+          sed -i '/python.toolchain(/a\    ignore_root_user_error = True,' MODULE.bazel
       - name: Patch source files
+        # these patches will be removed when the repo is updated/tested accordingly
```
charleshofer: Which repo? Right now this whole step is pretty unreadable, so it'd be good to eliminate if possible

PR #151, opened 2025-10-21
Enable RBE for builds and test

charleshofer:
## Motivation

Add the proper Bazel configuration to do remote builds with Bazel's RBE feature. This should greatly improve build times in CI.

## Test Plan

Ran the `build/ci_build.py` script locally and confirmed that it'll produce wheels with ROCm 7

### DIFF
--- a/ci/jax_rbe/pr_test.sh
+++ b/ci/jax_rbe/pr_test.sh
@@ -43,5 +43,9 @@ python3 build/build.py build --wheels=jax-rocm-plugin --configure_only --python_
     --test_verbose_timeout_warnings \
     --test_output=errors \
     //tests:core_test_gpu \
-    //tests:linalg_test_gpu
+    //tests:linalg_test_gpu \
+    --test_filter=CoreTest \
+    --test_filter=JaxprTypeChecks \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -59,6 +62,11 @@ jobs:
           repository: rocm/jax
           ref: rocm-jaxlib-v0.7.1
           path: jax
+      - name: Apply patches to rocm/jax test repo
+        run: |
+          pushd jax
+          git apply ../ci/patches/*
+          popd
```
JehandadKhan: Shouldnt this be done automatically by bazel  ? Why do we need to do this manually ?

alekstheod (approved):

.github/workflows/ci.yml
charleshofer: Because these get applied to the rocm/jax that we use for test cases, not the jax-ml/jax we use for wheel builds. Maybe it'd be better to carry this as a change in the rocm/jax repo itself rather than a patch? I'd really rather not carry anything in that repo though, and it's unfortunate that we have to keep jaxlib changes in there right now.

PR #155, opened 2025-10-24
Add explicit permissions for nightly

charleshofer:
## Motivation

I thought that this might be related to our permissions problems with images not being pulled from GHCR, but that is not the case since images are now pulling just fine. Anyways, it's good practice for workflows to explicitly state the permissions that they need, so we should still add this.

## Technical Details

Adds permissions that the nightly build job needs so that we don't rely on GITHUB_TOKEN's defaults: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions


psanal35 (approved):
LGTM~

PR #165, opened 2025-11-03
Fixed buidling docker with released-rocm or from ci build

zahiqbal:
## Motivation
Corrected "get_rocm.py"  command in bash script either get from released rocm or from ci build.

## Submission Checklist


tools/docker_dev_setup.sh
```diff
@@ -29,8 +29,8 @@ dpkg-reconfigure --frontend noninteractive tzdata
 
 # Default values
 rocm_version="7.1.0"
-rocm_build_number="16623"
-rocm_job_name="compute-rocm-dkms-no-npi-hipclang"
+#rocm_build_number="16623"
```
charleshofer: Let's not leave commented out code in here. Delete if it's not needed anymore. People can just pass the build number and job name in as parameters.

charleshofer (changes requested):

charleshofer (approved):
This can be removed because we want to use the publicly released ROCm packages, right?

PR #171, opened 2025-11-07
JAX Wheel Tagging with version extra

subodh-dubey-amd:
## Motivation - ⁠[SWDEV-563344](https://ontrack-internal.amd.com/browse/SWDEV-563344)

[SWDEV-563344](https://ontrack-internal.amd.com/browse/SWDEV-563344)
 - [ROCm QA][TheRock] Build specific JAX whl are not available in artifactory

This PR adds an wheel tagging feature enable rocm version management of JAX ROCm plugin wheels in automated build pipelines. The feature allows replacing the original version in wheel filenames with ROCm version tags, which is essential for maintaining consistent versioning across different ROCm builds and integrating with existing S3 upload workflows that use `$PACKAGE_DIST_DIR`.

## Technical Details
  - Replacing the version component (index 1) with the ROCm tag

**Pattern Transformation:**
```
jax_rocm7_plugin-0.7.1-cp312-cp312-manylinux_2_28_x86_64.whl
→ jax_rocm7_plugin+0.7.1+7.1.0a20251107-cp312-cp312-manylinux_2_28_x86_64.whl

jax_rocm7_pjrt-0.7.1-cp312-cp312-manylinux_2_28_x86_64.whl
→ jax_rocm7_pjrt-0.7.1+7.1.0a20251107-cp312-cp312-manylinux_2_28_x86_64.whl

```

## Test Result

https://github.com/ROCm/TheRock/actions/runs/19171644933

## Submission Checklist

### DIFF
--- a/jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py
+++ b/jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py
@@ -12,24 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Setup script for JAX ROCm plugin package."""
+
 import importlib
 import os
 from setuptools import setup
 from setuptools.dist import Distribution
 
 __version__ = None
-rocm_version = 0  # placeholder
-project_name = f"jax-rocm{rocm_version}-plugin"
-package_name = f"jax_rocm{rocm_version}_plugin"
+rocm_version = 0  # placeholder  # pylint: disable=invalid-name
+project_name = f"jax-rocm{rocm_version}-plugin"  # pylint: disable=invalid-name
+package_name = f"jax_rocm{rocm_version}_plugin"  # pylint: disable=invalid-name
 
 # Extract ROCm version from the `ROCM_PATH` environment variable.
-default_rocm_path = "/opt/rocm"
+default_rocm_path = "/opt/rocm"  # pylint: disable=invalid-name
 rocm_path = os.getenv("ROCM_PATH", default_rocm_path)
 rocm_detected_version = rocm_path.split("-")[-1] if "-" in rocm_path else "unknown"
 rocm_tag = os.getenv("ROCM_VERSION_EXTRA")
 
 
 def load_version_module(pkg_path):
+    """Load version module from the given package path.
+
+    Args:
+        pkg_path: Path to the package containing version.py
+
+    Returns:
+        The loaded version module
+    """
     spec = importlib.util.spec_from_file_location(
         "version", os.path.join(pkg_path, "version.py")
     )
@@ -39,16 +49,25 @@ def load_version_module(pkg_path):
 
 
 _version_module = load_version_module(package_name)
-__version__ = _version_module._get_version_for_build() 
+__version__ = (
+    _version_module._get_version_for_build()  # pylint: disable=protected-access
+)
 if rocm_tag:
     __version__ = __version__ + "+rocm" + rocm_tag
-_cmdclass = _version_module._get_cmdclass(package_name)
+_cmdclass = _version_module._get_cmdclass(  # pylint: disable=protected-access
+    package_name
+)
 
 
 class BinaryDistribution(Distribution):
     """This class makes 'bdist_wheel' include an ABI tag on the wheel."""
 
     def has_ext_modules(self):
+        """Indicate that this distribution has extension modules.
+
+        Returns:
+            bool: Always True to include ABI tag
+        """
         return True
 
 
--- a/jax_rocm_plugin/pjrt/python/setup.py
+++ b/jax_rocm_plugin/pjrt/python/setup.py
@@ -12,23 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Setup script for JAX ROCm PJRT plugin package."""
+
 import importlib
 import os
 from setuptools import setup, find_namespace_packages
 
 __version__ = None
-rocm_version = 0  # placeholder
-project_name = f"jax-rocm{rocm_version}-pjrt"
-package_name = f"jax_plugins.xla_rocm{rocm_version}"
+rocm_version = 0  # placeholder  # pylint: disable=invalid-name
+project_name = f"jax-rocm{rocm_version}-pjrt"  # pylint: disable=invalid-name
+package_name = f"jax_plugins.xla_rocm{rocm_version}"  # pylint: disable=invalid-name
 
 # Extract ROCm version from the `ROCM_PATH` environment variable.
-default_rocm_path = "/opt/rocm"
+default_rocm_path = "/opt/rocm"  # pylint: disable=invalid-name
 rocm_path = os.getenv("ROCM_PATH", default_rocm_path)
 rocm_detected_version = rocm_path.split("-")[-1] if "-" in rocm_path else "unknown"
 rocm_tag = os.getenv("ROCM_VERSION_EXTRA")
 
 
 def load_version_module(pkg_path):
+    """Load version module from the given package path.
+
+    Args:
+        pkg_path: Path to the package containing version.py
+
+    Returns:
+        The loaded version module
+    """
     spec = importlib.util.spec_from_file_location(
         "version", os.path.join(pkg_path, "version.py")
     )
@@ -38,7 +48,9 @@ def load_version_module(pkg_path):
 
 
 _version_module = load_version_module(f"jax_plugins/xla_rocm{rocm_version}")
-__version__ = _version_module._get_version_for_build() 
+__version__ = (
+    _version_module._get_version_for_build()  # pylint: disable=protected-access
+)
 if rocm_tag:
     __version__ = __version__ + "+rocm" + rocm_tag
 


build/ci_build
```diff
@@ -63,6 +63,51 @@ def _get_commit_info_from_wheel():
     }
 
 
+def retag_wheels(rocm_tag, package_dist_dir):
+    """
+    Retag wheel files by replacing version with ROCm version.
+    Pattern transformation: jax_rocm7_plugin-0.7.1-cp312-cp312-manylinux_2_28_x86_64.whl 
+    → jax_rocm7_plugin-7.10.0a20251021-cp312-cp312-manylinux_2_28_x86_64.whl
```
charleshofer: We cannot just replace the plugin version with the ROCm version like this. For one, JAX checks the plugin version against the JAX version when loading the plugin and this will confuse it. Two, there will be a bunch of plugin wheels out there that are confusingly labelled with versions that mean completely different things. Maybe a savvy developer will be able to tell that if the version is >7ish, then this version refers to the ROCm version that the plugin was built against. Three, we may now know what version of ROCm we built with, but we have no idea what version of the plugin this wheel was built from and what versions of JAX that this is compatible with.

There's almost certainly a better solution for encoding the ROCm version and build number in the wheel. What about the build tag section of the wheel's name? https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention

subodh-dubey-amd: https://github.com/ROCm/rocm-jax/pull/171/files#diff-b3bfb8bdea0e6e825c8c0d7132f89445e7d24535794f97e3cb3146b74a58095aR70

charleshofer:
> > 2. stalling via `pip` without pointing to the exact wheel URL a requirement, here?
> 
> [torch-2.10.0a0+rocm7.10.0a20251021-cp312-cp312-linux_x86_64.whl](https://rocm.nightlies.amd.com/v2-staging/gfx1151/torch-2.10.0a0%2Brocm7.10.0a20251021-cp312-cp312-linux_x86_64.whl) Pytorch follows the same pattern - I feel it comes from the upstream repo as version added

This is different. PyTorch is baking it into the version, notice the `+` rather than the `-`. I think the right solution is to bake the ROCm version into the plugin version, and then modify JAX's plugin handling code so that it'll ignore that ROCm bit after the `+`. I'm going to ask the rest of the team about what to do, but if you want to get moving on it you could start by changing things here:
https://github.com/ROCm/rocm-jax/blob/master/jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py#L40
https://github.com/ROCm/rocm-jax/blob/master/jax_rocm_plugin/pjrt/python/setup.py#L37

There should be a `__version__` variable that gets passed to the call to `setup()`. We would just need to append the ROCm version to it separate by a `+`, and that should probably be a configurable thing in the build.

subodh-dubey-amd:
> Discussing this more with JAX team: changing our version number to include `+rocm.1.2.3abc` isn't something that I think we should be doing. Changing the versioning scheme is a big deal. It will for sure require changes to the way that the plugin gets loaded and probably create a bunch of other bugs that we'll be dealing for a long time. Our CI is set up in such a way as to not have this problem with the ROCm version, and it sounds like this problem is specific to the way that TheRock's CI is set up (wanting to keep all the wheels in one big repository).
> 
> Scott suggested using the build variant feature that uv has (but you should make sure that it won't bust pip). And you can also always just download the wheel with something like wget or curl and install by giving pip a local path, too.
> 
> It seems like TheRock's CI was set up with PyTorch in mind, and we're not PyTorch. Encoding dependencies in the wheel name is a little odd to begin with, so is there a solution for you guys that doesn't involve that?



> Thanks **[charleshofer](/charleshofer)**

- Are you coordinating with the JAX team on this?

build/ci_build
```diff
@@ -299,6 +299,12 @@ def parse_args():
         help="Compiler backend to use when compiling jax/jaxlib",
     )
 
+    p.add_argument(
```
charleshofer: Please remove this. This argument isn't used anymore.

jax_rocm_plugin/pjrt/tools/build_utils.py
```diff
@@ -138,7 +138,7 @@ def update_setup_with_rocm_version(file_dir: pathlib.Path, rocm_version: str):
     with open(src_file, encoding="utf-8") as f:
         content = f.read()
     content = content.replace(
-        "rocm_version = 0  # placeholder", f"rocm_version = {rocm_version}"
+        "rocm_version = 0  # placeholder", f"rocm_version = \"{rocm_version}\""
```
charleshofer: Why are we adding double-quotes here?

jax_rocm_plugin/jaxlib_ext/tools/build_utils.py
```diff
@@ -134,6 +134,7 @@ def update_setup_with_cuda_version(file_dir: pathlib.Path, cuda_version: str):
 
 def update_setup_with_rocm_version(file_dir: pathlib.Path, rocm_version: str):
     """Update setup.py with the specified ROCm version."""
+
```
charleshofer: Remove this extra line

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -181,9 +181,13 @@ def build_jaxlib_wheel(
     cpy = to_cpy_ver(python_version)
     py_bin = "/opt/python/%s-%s/bin" % (cpy, cpy)
 
+
+    rocm_tag = os.environ.get("ROCM_VERSION", "7.1.0a20251030")
+
     env = dict(os.environ)
     env["JAX_RELEASE"] = str(1)
-    env["JAXLIB_RELEASE"] = str(1)
+    env["JAXLIB_RELEASE"] = f"{str(1)}+{rocm_tag}"
```
charleshofer: The JAXLIB_RELEASE flag doesn't do what you think it does. It's not a version number. This is a flag that gets set to either `1` or `0` depending on whether you want to build a "release" version of the wheel or a "dev" version. Remove the addition of the rocm_tag.

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -181,9 +181,13 @@ def build_jaxlib_wheel(
...
+    env["ROCM_VERSION"] = rocm_tag
```
charleshofer: We already pass a rocm_version argument into build/build.py (line 164). We don't want to also pass in an environment variable with the same name. That is very confusing. If the rocm_version argument give to build/build.py doesn't like the extra +rocm1.2.3, lets modify the scripts so that rocm_version will accept that rather than making an entire new environment variable for the ROCm version.

jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py
```diff
@@ -21,6 +21,7 @@
 rocm_version = 0  # placeholder
 project_name = f"jax-rocm{rocm_version}-plugin"
 package_name = f"jax_rocm{rocm_version}_plugin"
+rocm_tag = os.environ.get("ROCM_VERSION", "7.1.0a20251030")
```
charleshofer: We already have a section right below this that gets the ROCm version from the install path. Does the full ROCm version (alpha included) get written to /opt/rocm/.info/version? If so, lets read the ROCm version from there.

jax_rocm_plugin/pjrt/python/setup.py
```diff
@@ -34,7 +35,8 @@ def load_version_module(pkg_path):
   return module
 
 _version_module = load_version_module(f"jax_plugins/xla_rocm{rocm_version}")
-__version__ = _version_module._get_version_for_build()
+
```
charleshofer: Remove extra line

charleshofer:
This change should be retargeted to the master branch, not the 0.71. If this is going to make it into JAX 0.8.0/ROCm 7.10, it needs to be there, not here.

jax_rocm_plugin/pjrt/python/version.py
```diff
@@ -77,15 +77,19 @@ def _get_version_for_build() -> str:
     The returned version string depends on which environment variables are set:
     - if JAX_RELEASE or JAXLIB_RELEASE are set: version looks like "0.4.16"
     - if JAX_NIGHTLY or JAXLIB_NIGHTLY are set: version looks like "0.4.16.dev20230906"
-    - if none are set: version looks like "0.4.16.dev20230906+ge58560fdc
+    - if none are set: version looks like "0.4.16.dev20230906+ge58560fdc"
+    - if ROCM_VERSION is set: adds ROCm build tag like "0.4.16+rocm7.1.0a20251107"
```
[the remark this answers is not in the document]
subodh-dubey-amd: Removed!

jax_rocm_plugin/build/rocm/tools/build_wheels.py
subodh-dubey-amd: REMOVED!

jax_rocm_plugin/pjrt/tools/build_utils.py
subodh-dubey-amd: This was just for testing purposes. It’s outdated now, so I’ve removed it

jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py
subodh-dubey-amd: This was just for testing purposes. It’s outdated now, so I’ve removed it

jax_rocm_plugin/pjrt/python/setup.py
subodh-dubey-amd: Removed!

jax_rocm_plugin/pjrt/python/setup.py
```diff
@@ -20,6 +20,7 @@
 rocm_version = 0  # placeholder
 project_name = f"jax-rocm{rocm_version}-pjrt"
 package_name = f"jax_plugins.xla_rocm{rocm_version}"
+rocm_tag = os.environ.get("ROCM_VERSION", "7.1.0a20251030")
```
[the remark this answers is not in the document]
subodh-dubey-amd: used **rocm_detected_version**

jax_rocm_plugin/build/rocm/tools/build_wheels.py
subodh-dubey-amd: This was just for testing purposes. It’s outdated now, so I’ve removed it

jax_rocm_plugin/pjrt/python/setup.py
```diff
@@ -24,17 +24,20 @@
...
-  module = importlib.util.module_from_spec(spec)
-  spec.loader.exec_module(module)
-  return module
+    spec = importlib.util.spec_from_file_location(
+        "version", os.path.join(pkg_path, "version.py")
+    )
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
```
subodh-dubey-amd: changed this because linting (black) was failing

subodh-dubey-amd:
Build Portable Linux JAX Wheels #470 - https://github.com/ROCm/TheRock/actions/runs/19292401110/job/55166014993

subodh-dubey-amd:
> This change should be retargeted to the master branch, not the 0.71. If this is going to make it into JAX 0.8.0/ROCm 7.10, it needs to be there, not here.

Changed the base branch to main

jax_rocm_plugin/pjrt/python/setup.py
```diff
@@ -24,17 +24,20 @@
 # Extract ROCm version from the `ROCM_PATH` environment variable.
 default_rocm_path = "/opt/rocm"
 rocm_path = os.getenv("ROCM_PATH", default_rocm_path)
-rocm_detected_version = rocm_path.split('-')[-1] if '-' in rocm_path else "unknown"
+rocm_detected_version = rocm_path.split("-")[-1] if "-" in rocm_path else "unknown"
+rocm_tag = os.getenv("ROCM_VERSION_EXTRA", rocm_detected_version)
```
mminutoli: Same issue would happen here.

subodh-dubey-amd: Yes, but we have passed the 
`"-e",
            "ROCM_VERSION_EXTRA=" + rocm_version,`

jax_rocm_plugin/build/rocm/ci_build

So, it should never be empty

jax_rocm_plugin/jax_plugins/rocm/plugin_setup.py
```diff
@@ -26,6 +26,7 @@
 default_rocm_path = "/opt/rocm"
 rocm_path = os.getenv("ROCM_PATH", default_rocm_path)
 rocm_detected_version = rocm_path.split("-")[-1] if "-" in rocm_path else "unknown"
+rocm_tag = os.getenv("ROCM_VERSION_EXTRA")
```
mminutoli: Sorry, I might be pedantic on my side but I don't see this as an improvement. Now if `ROCM_VERSION_EXTRA` is not set rocm_tag is `""` and the version will include just `+rocm` without the ROCm version which I don't think it is the intended behavior. What we need here is a fallback mechanism that is guaranteed to give us the ROCm version with which we are building the artifact when the environment variable is not set. Maybe using `rocminfo`?

mminutoli (changes requested):

charleshofer (approved):

PR #174, opened 2025-11-13
Add symbols to wheel builds in the CI

JehandadKhan:
## Motivation

In many instances it is useful to have symbols in the wheel to facilitate triage etc. 

## Technical Details

When the `build.py` script is triggered, we add the appropriate bazel flags to retain symbols in the different SO files in our wheels.

## Test Plan

The produced artifacts are manually tests using the `nm -D` command to ensure that symbols are present.

## Test Result

The output of the `nm -D ` command shows that the symbols are indeed present in all the SOs in the wheel

## Submission Checklist

- [X ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


tools/docker_dev_setup.sh
```diff
@@ -118,7 +118,7 @@ apt-get install -y \
   build-essential \
   make \
   patchelf \
-  python3.10-venv \
+  python3-venv \
```
charleshofer: Good catch

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -159,6 +159,9 @@ def build_jaxlib_wheel(
         "python",
         "build/build.py",
         "build",
+        "--bazel_options=--strip=never",
```
charleshofer: I'd prefer that as many Bazel options as possible make their way down to the `jax_rocm_plugin/.bazelrc` file rather than getting threaded through one of our dozen-or-so wrapper scripts. Preferably, it'd be inside of something like a `build:debug` config, so you can enable it with `./bazel build ... --config=debug ...` or `build/buid.py ... --bazel_options="--config=debug" ...`.

charleshofer (changes requested):
You might want to talk to Aleksei about how this will relate to the debug flags he put in the Makefile template. We it'd be nice if we could unify on one set of debug flags that can just sit in `.bazelrc` under their own config: https://github.com/ROCm/rocm-jax/blame/master/stack.py#L28

jax_rocm_plugin/build/rocm/tools/build_wheels.py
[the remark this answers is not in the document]
charleshofer: But yes, we should use that instead, and modify it with whatever flags we want

jax_rocm_plugin/build/rocm/tools/build_wheels.py
```diff
@@ -159,6 +159,7 @@ def build_jaxlib_wheel(
         "python",
         "build/build.py",
         "build",
+        "--bazel_options=--config=debug_symbols",
```
charleshofer: Can we stick this in the `build/build.py` instead rather than storing bazel arguments directly in here?

JehandadKhan:
@Ruturaj4  and @charleshofer  I believe I have addressed the required changes, can you please review. If the comments above are addressed please mark them as resolved

PR #189, opened 2025-11-20
Add RBE secrets to nightly

charleshofer:
https://github.com/ROCm/rocm-jax/pull/151 enabled RBE for PR builds and changed the build-wheels.yml to require secrets to be passed into the workflow. However, 151 didn't adjust the nightly workflow to pass the required secrets when it calls the wheel build workflow. This change adds those required secrets to the workflow call.


psanal35 (approved):
LGTM~

PR #198, opened 2025-11-27
Fix symlink command from ENV to RUN

Ruturaj4:
## Motivation

Problem is that the dockerfile is using ENV instead of RUN for symlink creation command.

## Test Plan

Tested by creating a new docker image.

## Submission Checklist

### DIFF
--- a/docker/Dockerfile.jax-ubu24
+++ b/docker/Dockerfile.jax-ubu24
@@ -30,8 +30,8 @@ RUN --mount=type=bind,source=jax_rocm_plugin/build/rocm/tools/get_rocm.py,target
     --mount=type=cache,target=/var/cache/apt \
     --mount=type=bind,from=therock,target=/tmp/therock/ \
     python3 get_rocm.py --rocm-version=$ROCM_VERSION --job-name=$ROCM_BUILD_JOB --build-num=$ROCM_BUILD_NUM --therock-path=$THEROCK_PATH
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.jax-ubu24
```diff
@@ -30,8 +30,8 @@ RUN --mount=type=bind,source=jax_rocm_plugin/build/rocm/tools/get_rocm.py,target
     --mount=type=cache,target=/var/cache/apt \
     --mount=type=bind,from=therock,target=/tmp/therock/ \
     python3 get_rocm.py --rocm-version=$ROCM_VERSION --job-name=$ROCM_BUILD_JOB --build-num=$ROCM_BUILD_NUM --therock-path=$THEROCK_PATH
+RUN ln -s "/opt/rocm-${ROCM_RELEASE}/lib/librocblas.so.5" "/opt/rocm-${ROCM_RELEASE}/lib/librocblas.so.4"
```
mminutoli: Do you know when this is needed?! Can we just remove it?

Ruturaj4: yes I think on latest rocms like `rocm7.2.0` but this needs a proper fix tbh

draganmladjenovic: It is a dangerous hack. And should not be used.

Ruturaj4:
@charleshofer can you also take a look?

charleshofer:
Yeah, this was a hack to fix issues in ROCm a few releases ago. We should be able to just remove this since the link isn't even being created.

PR #201, opened 2025-11-30
Refactor commit_info.py to use single 'commits' dict

AratiGanesh:
Refactors commit_info.py to use a single commits dictionary instead of separate variables, making it simpler to access. 

Usage - 
from jax_plugins.xla_rocm7.commit_info import commits
print(commits)


Arech8 (approved):
Great, thanks a ton, @AratiGanesh !

PR #216, opened 2025-12-10
Bump commit for profiling changes on 0.8.0

charleshofer:
Bump the XLA commit to pick up rocprofiler-sdk fixes in https://github.com/ROCm/xla/pull/473


i-chaochen:
do we still have any further changes to 0.8.0? @alekstheod It's ok to me now unless you have more changes need to add into 0.8.0

PR #226, opened 2025-12-17
Improve check for AMD GPU availability; drop the patchfile

Arech8:
## Motivation

Previous check invoked `hipInit()` which led to rocprofiler-sdk misconfiguration

## Technical Details

This check tests presence of certain KFD files describing properties of compute units and determines if GPUs are there.

## Test Plan

Manual testing.

### DIFF
--- a/jax_rocm_plugin/pjrt/python/__init__.py
+++ b/jax_rocm_plugin/pjrt/python/__init__.py
@@ -195,7 +195,7 @@ def initialize():
         return
 
     if not is_amd_gpu_available():
-        raise ValueError("No GPUs found")
+        raise ValueError("No AMD GPUs were found, skipping ROCm plugin initialization")
 
     options = xla_client.generate_pjrt_gpu_plugin_options()
     options["platform_name"] = "ROCM"


jax_rocm_plugin/pjrt/python/__init__.py
```diff
@@ -128,6 +130,54 @@ def set_rocm_paths(path):  # pylint: disable=too-many-branches
...
+                file_size = os.path.getsize(node_props_path)
+                # 16KB is more than a reasonable limit
+                if file_size <= 0 or file_size > 16 * 1024:
+                    continue
+
+                with open(node_props_path, "r", encoding="ascii") as f:
+                    match = r_wave_front_size.search(f.read())
+                    if match:
+                        wave_front_size = int(match.group(1))
+                        if wave_front_size > 0:
```
charleshofer: I'm guessing that we check `wave_front_size` to filter out AMD CPUs? What else would have a `wave_front_size` of 0?

jax_rocm_plugin/pjrt/python/__init__.py
```diff
@@ -128,6 +130,54 @@ def set_rocm_paths(path):  # pylint: disable=too-many-branches
...
+    code. And we don't want to initialize HIP right now inside the current
+    process, because doing so might spoil a proper initialization of the
+    rocprofiler-sdk later during PJRT startup."""
+
+    try:
+        kfd_nodes_path = "/sys/class/kfd/kfd/topology/nodes/"
+        if not os.path.exists(kfd_nodes_path):
+            return False
+
+        r_wave_front_size = re.compile(r"\bwave_front_size\s+(\d+)\b", re.MULTILINE)
```
charleshofer: nit: It'd be nice to have a comment explaining what we expect this regex is for and what it'll match.

jax_rocm_plugin/pjrt/python/__init__.py
```diff
@@ -128,6 +130,54 @@ def set_rocm_paths(path):  # pylint: disable=too-many-branches
     os.environ["JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH"] = lld_path
```
draganmladjenovic: We should remove this at this point. I'm not saying that should be part of this change.

jax_rocm_plugin/pjrt/python/__init__.py
```diff
@@ -140,8 +190,7 @@ def initialize():
         logger.warning("rocm_plugin_extension not found")
         return
 
-    device_count = rocm_plugin_extension.get_device_count()
-    if device_count <= 0:
+    if not is_amd_gpu_available():
         raise ValueError("No GPUs found")
```
draganmladjenovic: Do we know why do we raise here instead of loging + return as above?

PR #228, opened 2025-12-17
Add dev container workflow

mminutoli:
## Motivation

Our CI rebuilds docker containers in many workflows. While the step takes relatively short time to execute, most of the work is redundant and can be cached in a base image.  The base image can also serve as starting point to provide a pre-canned development environment with all dependencies already installed as it's been requested several time by the teams we collaborate with.

## Technical Details

The patch is still in development and missing substantial pieces. At the current stage I am interested in getting early feedback as I want to avoid breaking "contracts" that we might have with other teams.

When the work will be completed the set of patches will include a CI workflow that periodically (e.g., once a week) will generate a set of containers from scratch and store them in a registry. Some will have LLVM pre-installed with the intent of being pull ready dev-environments and others will be without it to keep the container leaner. The base containers are created by the `build_base_dockers` that I have added to the `ci_build`.

Currently, the base image that is referenced in the `Dockerfile.jax-{ubu22,ubu24}` are wired to use the image for ROCm 7.1.1 and need updating once the other pieces are in place. 
 
## Test Plan

I have currently tested the addition and the previous commands manually for ROCm 7.1.1, but the patch set will be ready when all CI workflows will work as expected.

## Test Result

The tests that I have done manually have successfully built base images and the docker containers that the CI uses.

## Submission Checklist

### DIFF
--- a/build/ci_build
+++ b/build/ci_build
@@ -53,6 +53,11 @@ def _check_base_image_exists(base_image_name):
         return False
 
 
+def _is_github_actions():
+    """Check if the script is running in a GitHub Actions environment."""
+    return os.environ.get("GITHUB_ACTIONS") == "true"
+
+
 def _tag_and_push_latest(local_tag, registry="ghcr.io/rocm"):
     """
     Tag a local Docker image and push it to GHCR with 'latest' tag.
@@ -60,6 +65,9 @@ def _tag_and_push_latest(local_tag, registry="ghcr.io/rocm"):
     This function tags a locally built image and pushes it to GitHub Container
     Registry with the 'latest' tag, following the same naming scheme as the
     build-base-docker.yml workflow.
+
+    Note: Pushing to GHCR only occurs when running in a GitHub Actions environment.
+    When running locally, the image is tagged but not pushed.
 
     Args:
         local_tag (str): Local Docker image tag (e.g., "jax-base-ubu22.rocm711")
@@ -71,8 +79,12 @@ def _tag_and_push_latest(local_tag, registry="ghcr.io/rocm"):
     ghcr_image = "%s/%s:latest" % (registry, local_tag)
     print("Tagging %s as %s" % (local_tag, ghcr_image))
     subprocess.check_call(["docker", "tag", local_tag, ghcr_image])
-    print("Pushing %s" % ghcr_image)
-    subprocess.check_call(["docker", "push", ghcr_image])
+    
+    if _is_github_actions():
+        print("Pushing %s" % ghcr_image)
+        subprocess.check_call(["docker", "push", ghcr_image])
+    else:
+        print("Skipping push to GHCR (not running in GitHub Actions)")
 
 
 def _get_commit_info_from_wheel():


build/ci_build
```diff
@@ -121,35 +121,166 @@ def _fetch_jax_metadata(xla_path):
...
+        temp_path.mkdir(exist_ok=True)
+        cmd.append("--build-context=therock=%s" % temp_path)
+
+    # context dir
+    cmd.append(".")
+
+    subprocess.check_call(cmd)
+
+
+def build_dockers(
```
charleshofer: This PR effectively changes our build process for JAX from,

1. build/ci_build ... dist_wheels
2. build/ci_build ... build_dockers
3. build/ci_build ... test

to

1. build/ci_build ... dist_wheels
2. build/ci_build ... build_base_dockers
3. build/ci_build ... build_dockers
4. build/ci_build ... test

Changing the commands that we use for building is tricky since this process is a sort of contract between us and the ROCm DevOps team. The ROCm DevOps team depends on the build process being stable, and they expect to be able to plug in their nightly builds of ROCm or TheRock to the build process in order to test ROCm against JAX. Getting them to change their CI workflows is hard. I'm not saying that we shouldn't move to the base docker model, but I think we should try and do so in a way that the process doesn't have to change for them. Maybe add an `if` somewhere in the `build_dockers` code path so that it will build the base image if the ROCm version can't be found or if the build number and build job are different?

mminutoli: No, it should stay the same with my last changes. GitHub registry will serve the base prebuilt image to the dockerfile used by `build_dockers`.

mminutoli:
> I wonder how does this different to upstream jax and xla dev container? https://github.com/jax-ml/jax/blob/main/ci/utilities/README.md and https://github.com/openxla/xla/blob/03ffe4eb56a51640c990bb522f72926ac80e5c66/.github/workflows/ci.yml#L94-L109
> 
> They are using tensorflow as the dev container and xla followed the same logic. We provided it in our TF repo https://github.com/ROCm/tensorflow-upstream/tree/develop-upstream/ci/official

@i-chaochen likely at this point in time there is not much difference. In a future pull request, I'll start putting the containers on a diet as they are quite big right now. This patch set is currently just trying to avoid rebuilding the same docker images over and over *and* to provide a pre-canned dev environment with things like LLVM already installed so that it'll will be easier on your side when you'll need them.

.github/workflows/build-base-docker.yml
```diff
@@ -0,0 +1,102 @@
...
+      fail-fast: false
+      matrix:
+        ubuntu-version: ["22", "24"]
+        rocm-version: ["7.1.1", "7.2.0"]
+        install-llvm: [true, false]
+        include:
+          - rocm-version: "7.1.1"
+            runner-label: "mi-250"
+          - rocm-version: "7.2.0"
+            rocm-build-job: "compute-rocm-dkms-no-npi-hipclang"
```
charleshofer: Could we switch this to a release candidate build?

PR #239, opened 2025-12-30
[WIP] Introduce jax nightly multigpu ci script

alekstheod:
Introduce multigpu test script for rocm

### DIFF
--- a/jax_rocm_plugin/build/rocm/run_jax_multigpu_ut.sh
+++ b/jax_rocm_plugin/build/rocm/run_jax_multigpu_ut.sh
@@ -7,7 +7,6 @@ declare -A args
 args['--python_version']=""
 args['--jax_version']=""
 args['--jax_dir']=""
-args['--jax_plugin_dir']=""
 args['--xla_dir']=""
 
 while [ $# -gt 0 ]; do
@@ -22,29 +21,14 @@ while [ $# -gt 0 ]; do
     shift
 done
 
+SCRIPT_DIR=$(realpath $(dirname $0))
 JAX_DIR=${args['--jax_dir']}
-JAX_PLUGIN_DIR=${args['--jax_plugin_dir']}
 WHEELHOUSE="${JAX_DIR}/../wheelhouse"
 XLA_DIR=${args['--xla_dir']}
 
 clean_up() {
     rm -rf ${WHEELHOUSE}
 }
-
-trap clean_up EXIT
-
-pushd ${JAX_PLUGIN_DIR}
-python3 build/build.py build \
-    --use_clang=true \
-    --clang_path=/lib/llvm-18/bin/clang-18 \
-    --wheels=jax-rocm-plugin,jax-rocm-pjrt \
-    --target_cpu_features=native \
-    --rocm_path=/opt/rocm \
-    --rocm_version=7 \
-    --rocm_amdgpu_targets=${AMDGPU_TARGETS} \
-    --local_xla_path=${XLA_DIR} \
-    --output_path=${WHEELHOUSE} \
-    --verbose
 
 pushd ${JAX_DIR}
 JAX_VERSION=${args['--jax_version']}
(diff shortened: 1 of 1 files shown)


jax_rocm_plugin/build/rocm/run_jax_multigpu_ut.sh
```diff
@@ -0,0 +1,107 @@
...
+    --config=rocm_rbe \
+    --//jax:build_jaxlib=false \
+    --override_repository=xla=${XLA_DIR} \
+    --build_tag_filters=cpu,gpu,-tpu,-config-cuda-only \
+    --test_tag_filters=cpu,gpu,-tpu,-config-cuda-only \
+    --action_env=TF_ROCM_AMDGPU_TARGETS="gfx906,gfx908,gfx90a,gfx942,gfx950,gfx1030,gfx1100,gfx1101,gfx1200,gfx1201" \
+    --test_verbose_timeout_warnings \
+    --test_output=errors \
+    -- \
+    //tests:multiprocess_gpu_test \
```
charleshofer: Thanks for getting this list of tests together

jax_rocm_plugin/rbe.bazelrc
```diff
@@ -1,24 +1,24 @@
...
+build:rocm_rbe --tls_client_key="/tf/certificates/ci-cert.key"
 
 build:rocm_rbe --bes_backend="grpcs://wardite.cluster.engflow.com"
 build:rocm_rbe --bes_results_url="https://wardite.cluster.engflow.com/invocation/"
 build:rocm_rbe --remote_executor="grpcs://wardite.cluster.engflow.com"
 build:rocm_rbe --remote_cache="grpcs://wardite.cluster.engflow.com"
-build:rocm_rbe --host_platform="//platform/linux:manylinux"
-build:rocm_rbe --extra_execution_platforms="//platform/linux:manylinux"
-build:rocm_rbe --platforms="//platform/linux:manylinux"
+build:rocm_rbe --host_platform="@local_config_rocm//rocm:linux_x64"
```
charleshofer: Two things here,

1. I don't remember the details, but these were parameterized somehow so that we could pass in different docker images into the platforms right? Could we pass in the images from here, https://github.com/ROCm/rocm-jax/blob/master/jax_rocm_plugin/platform/linux/BUILD#L13?
2. Also, if we start using the platforms from the rocm/xla repo, we should remove our platforms in `jax_rocm_plugin/platorms` directory.

jax_rocm_plugin/build/rocm/run_jax_multigpu_ut.sh
```diff
@@ -0,0 +1,107 @@
...
+    //tests:aot_test_gpu \
+    //tests:mock_gpu_topology_test_gpu \
+    //tests:ann_test_gpu \
+    //tests:debugging_primitives_test_gpu \
+    //tests:array_test_gpu \
+    //tests:export_test_gpu \
+    //tests:memories_test_gpu \
+    //tests:debugger_test_gpu \
+    //tests:lax_control_flow_test_gpu \
+    //tests:checkify_test_gpu
```
charleshofer: I've looked into it, but haven't been able to find a solution to this, but it'd be awesome if we could skip or deselect individual test cases (like [tests/linalg_test.py::NumpyLinAlg::testCholeskyGradPrecision](https://github.com/ROCm/jax/blob/main/tests/linalg_test.py#L121C7-L121C32)) rather than whole files by passing them into this command somehow.

I've tried with the `--test_filter` option to no avail. It maps to unittest's `-k` option, which is not very powerful.

Not necessary, so no worries if there isn't a way or you can't find one. We might just be limited by the fact that Bazel is using unittest/absl under the hood.

jax_rocm_plugin/rbe.bazelrc
alekstheod: I believe removing the platforms shall be a part of a separate PR

jax_rocm_plugin/build/rocm/run_jax_multigpu_ut.sh
alekstheod: I found this:
```
--test_arg=-k \
--test_arg='not test_fft_large and not test_conv3d and not test_xmap'
```
Will try it out...

PR #243, opened 2026-01-08
To build and install jax, jaxlib, pjrt and plugin at once.

i-chaochen:
## Motivation

Now we can just use one line to build everything (jax, jaxlib, pjrt and plugin) as we're used to.

```
python3 stack.py build --xla-dir=/my/own/xla/path --jax-dir=/my/own/jax/path
```


charleshofer (approved):
Looks fine to me

Arech8:
Also why do you make two separate PRs (into the `master` and here), instead of having just one main PR merged into the `master` and then cherry-picking it here? I think, this PR should be discarded in favor of cherry-picking, this helps with maintenance.

PR #246, opened 2026-01-09
[WIP] Stack direct build without Makefile

i-chaochen:
This is a more aggressive change based on https://github.com/ROCm/rocm-jax/pull/243 

The build is simple by one line without `Makefile`

```
python3 stack.py build --xla-dir=/your/xla --jax-dir=/your/jax
```

1. Direct Build Implementation: Refactored `stack.py` to handle the entire build and installation process (JAX, jaxlib, PJRT, and plugin) directly using Python subprocesses, removing the dependency on a `Makefile`.

2. Simplified Command Line:
   - Renamed `--kernel-jax-dir` to `--jax-dir`.
   - Enabled the use of the = syntax (e.g., `--xla-dir=/path/to/xla`).
   - Made `--rebuild-makefile` implicit for the build action.

3. New flag `--debug`: You can now add `--debug` to your build or develop commands.
 - Debug configurations: When `--debug` is used, the build process will:
     - Enable `--config=debug` and `--compilation_mode=dbg`
     - Disable stripping of symbols (`--strip=never`).
     - Set maximum debug info level (`-g3`).
     - Disable optimizations (`-O0`) for both C and C++ code.
  - Makefile support: Running `python3 stack.py develop --debug` will generate a `Makefile` configured to use the `CFG_DEBUG` options by default.
  - Example usage:
      - To build everything in debug mode:
```
python3 stack.py build --xla-dir=/your/xla --jax-dir=/your/jax --debug
```
To just generate a Makefile with debug settings:
```
python3 stack.py develop --xla-dir=/your/xla --jax-dir=/your/jax --debug
```

If you really want `Makefile` you can still go with 

```
python3 stack.py develop --xla-dir=/your/xla --jax-dir=/your/jax
```

4. Specific clang path and choose clang18 as default one

```
python3 stack.py build --xla-dir=/your/xla --jax-dir=/your/jax --clang-path=/usr/lib/llvm-18/bin/clang
```

### DIFF
--- a/stack.py
+++ b/stack.py
@@ -493,9 +493,7 @@ def build_and_install(
 
     this_repo_root, xla_path, jax_path = _resolve_relative_paths(xla_dir, jax_dir)
     amdgpu_targets = get_amdgpu_targets()
-
-    # Use "system" to force build to use the container's python version
-    python_version = "system"
+    python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
 
     if not clang_path:
         clang_path = find_clang() or "/usr/lib/llvm-18/bin/clang"
--- a/stack.py
+++ b/stack.py
@@ -137,29 +137,37 @@ refresh_jaxlib: jaxlib_clean jaxlib jaxlib_install
 def find_clang():
     """Find a local clang compiler and return its file path."""
 
-    clang_path = None
+    # 1. Prioritize specific LLVM versions
+    llvm_paths = [
+        "/usr/lib/llvm-18/bin/clang"
+    ]
+    for path in llvm_paths:
+        if os.path.exists(path):
+            return path
 
-    # check PATH
+    # 2. Check PATH
     try:
         out = subprocess.check_output(["which", "clang"])
-        clang_path = out.decode("utf-8").strip()
-        return clang_path
+        return out.decode("utf-8").strip()
     except subprocess.CalledProcessError:
         pass
 
-    # search /usr/lib/
+    # 3. search /usr/lib/llvm* directories as a fallback
     top = "/usr/lib"
-    for root, dirs, files in os.walk(top):
-        # only walk llvm dirs
-        if root == top:
-            for d in dirs:
-                if not d.startswith("llvm"):
-                    dirs.remove(d)
+    if os.path.exists(top):
+        for root, dirs, files in os.walk(top):
+            # only walk llvm dirs
+            if root == top:
+                # Prioritize higher versions by sorting reverse
+                dirs.sort(reverse=True)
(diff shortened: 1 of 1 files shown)


Ruturaj4 (approved):
LGTM @charleshofer can you also take a look?

Arech8 (changes requested):
I don't think adding another build sequence to support is useful and valuable. `Makefile` route works flawlessly, why should we add another orthogonal way to build things? This creates nothing, but a maintenance hell...
 
IMO the PR should be discarded.

i-chaochen:
> Have you read @charleshofer descriptions of the process? The old process is as simple as doing:

which docs? not in https://github.com/ROCm/rocm-jax/blob/master/BUILDING.md and 

https://github.com/ROCm/rocm-jax/blob/master/DEVSETUP.md#setting-up-a-docker-development-environment

is let me to do this thing

```
python3 stack.py develop --rebuild-makefile
cat jax_rocm_plugin/Makefile
```
so I edit my xla and jax path with what docs mentioned here

```
source .venv/bin/activate
make clean dist install
```

there are no use cases in the docs to what you said

> 1. `python3 stack.py develop` with correct flags once (flags to specify JAX and XLA existed there almost all the time, you must have seen them when modifying the file, didn't you?), + `cd $YOUR_JAX_DIR && pip install . -e once`.
> 2. `make refresh refresh_jaxlib` each time you need to install/update you jax instance.

Arech8:
> so you mean we can just do like this?
> python3 stack.py develop xla_dir=/my/own/xla/ test_jax_ref=/my/own/jax ?
> BTW, why you got test_jax_ref and kernels_jax_dir ? cannot we just to have one only? is test_jax_ref is for jax repo and kernels_jax_dir is for jaxlib?

@i-chaochen  can you please finally have a look at stack.py CLI (that you proposed to modify, meaning you *must* already understand them) at https://github.com/ROCm/rocm-jax/blob/master/stack.py#L398 or just run `stack.py -h` and read the CLI arguments and their meaning? It's just a couple of dozen lines of simple readable code that contain all the answers. You can find correct arguments on your own, I believe in you!

> I would like to add an use-case for ...

sure, please make a PR

i-chaochen:
> @i-chaochen please don't waste an effort on this PR, we will not merge a third parallel way to build the repo.
> 
> On top of that, as was said, @alekstheod is brewing a major refactoring of the build system to bazelize things as much as possible, so this will be obsolete in a very short time.

my understanding is @alekstheod's work is about RBE in CI, I am not sure how it will be. For now, I'm using this for my team purpose to build.

PR #257, opened 2026-01-15
Run single and multi GPU tests in parallel and upload combined results

psanal35:
Previously, each uploaded log represented a full test suite run (single+multi GPU). To preserve this model while enabling parallel execution, single and multi GPU tests are now run as separate parallel jobs and their results are combined before uploading.


psanal35:
This feature is no longer needed since the current test suite is already updated.

PR #270, opened 2026-01-22
add hardcoded or fixed number to fetch rocm 7.2.0

Ruturaj4:
## Motivation

We can't use "lastsuccessful" run of the `compute-rocm-dkms-no-npi-hipclang` as they tend to change the version of rocm over time. Current "lastsuccessful" points to the rocm 7.3.0. To workaround this we still need to hardcode code this, and keep on using this until the rock migration is complete.

image


charleshofer (changes requested):

psanal35 (approved):

charleshofer (changes requested):
Now that 7.2.0 is public, we shouldn't need this PR anymore. You can probably just close it.

PR #279, opened 2026-01-23
Update model performance workflows with ROCm 7.2.0 public release

psanal35:
Workflows updated to use the public ROCm 7.2.0 release too (See PR https://github.com/ROCm/rocm-jax/pull/271). Similarly, performance models will also be updated.


charleshofer (approved):
Thanks for updating these

psanal35:
It was addressed in another PR.

PR #288, opened 2026-01-30
Fix metrics parsing in Llama custom workload and DB upload

psanal35:
This PR fixes the custom workload metrics parsing so training outputs are correctly sent into the database when datasets change. It is no longer relies on a fixed dataset prefix. It also now use explicit CLI arguments for DB uploads instead of hardcoded values and implicit environment assumptions.

PS. MoE is disabled in CI as part of this PR

### DIFF
--- a/ci/upload_to_llama_db.py
+++ b/ci/upload_to_llama_db.py
@@ -1,9 +1,9 @@
 """Upload Llama training summary results to MySQL database."""
 
+import os
 import argparse
 import ast
 import json
-import os
 from datetime import date
 
 # pylint: disable=import-error


ci/upload_to_llama_db.py
```diff
@@ -1,8 +1,9 @@
 """Upload Llama training summary results to MySQL database."""
 
-import os
+import argparse
 import ast
 import json
+import os
```
magaonka-amd: cosmetic suggestion: why remove import os and add it here again. I suggest you leave it where it was before.

PR #290, opened 2026-02-03
Use wheel 0.46.3 in fixwheel.py

charleshofer:
There's a [security vulnerability](https://www.sentinelone.com/vulnerability-database/cve-2026-24049/) in `wheel==0.46.2`. This change moves us up to version 0.46.3 or higher to get the fix for it. However, the `wheel.cli` Python interface was removed in version 0.46.3. So, also convert the places where `wheel` is used to use the CLI via `subprocess` instead of the Python interface.


ScottTodd (approved):

Arech8:
@charleshofer , for the future, you can just silence the linter with a comment `# pytype: disable=import-error,import-not-found`. Having an additional comment why this is needed ("installed during the build process and not necessary available to linter" is always a great idea for maintainability.

PR #299, opened 2026-02-10
[WIP] Remove patchelf usage

alekstheod:
Remove usage patchelf and use linkopts instead to set the rpaths.
This PR is a collection of changes that are required to be done in order to avoid using 
external tools to control the rpaths on a specific target while building xla/jax.

There is upstreamed PR for xla that adds the basic support: https://github.com/openxla/xla/pull/37513

If we want to use this approach we can 
1 - land the changes to our xla release branch
2 - depending on how we want to go further
     * set rpaths directly in jax targets (upstream jax/release branches)
     * or move jax targets into rocm-jax and set the linkopts there!

### DIFF
--- a/jax_rocm_plugin/build/build.py
+++ b/jax_rocm_plugin/build/build.py
@@ -611,6 +611,7 @@ async def main():
 
     if "rocm" in args.wheels:
         wheel_build_command_base.append("--config=rocm_base")
+        wheel_build_command_base.append("--config=rocm_wheel")
         if args.use_clang:
             wheel_build_command_base.append("--config=rocm")
             wheel_build_command_base.append(
--- a/jax_rocm_plugin/jaxlib_ext/rocm/rocm_plugin_extension.cc
+++ b/jax_rocm_plugin/jaxlib_ext/rocm/rocm_plugin_extension.cc
@@ -1,0 +1,126 @@
+/* Copyright 2024 The JAX Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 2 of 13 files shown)


Ruturaj4 (approved):

jax_rocm_plugin/build/build.py
```diff
@@ -611,6 +611,7 @@ async def main():
 
     if "rocm" in args.wheels:
         wheel_build_command_base.append("--config=rocm_base")
+        wheel_build_command_base.append("--config=rocm_wheel")
```
charleshofer: nit: This config just turns the -rpath settings on and off. Could we call this config something like `no_rocm_rpath`?

jax_rocm_plugin/jaxlib_ext/rocm/rocm_plugin_extension.cc
```diff
@@ -0,0 +1,126 @@
+/* Copyright 2024 The JAX Authors.
```
charleshofer: Similar to the BUILD file, it's going to be a pain to maintain this file unless we simply remove it from upstream. Is that the intent?

Also, can you explain what's different between this one and the upstream one?

charleshofer (changes requested):
I'm not really sure what the motivation is behind this PR. The description is kind of sparse. Can you explain the motivation for the PR and briefly explain how it works in the description?

This change does remove the rpath hack that we have in the Python scripts that do the wheel builds, but from what I can tell, the cost we're paying is copying a bunch of upstream Bazel and C++ code, adding another Bazel file specifically for handling rpaths, and adding yet another config option. Copying in rocm_plugin_extension.cc is especially tricky because upstream does modify that one a fair bit. Maybe there's reasons for bring BUILD and rocm_plugin_extension.cc into this repo and removing them from upstream? But right now, I'm inclined to think the cure is worth than the disease in this case.

Also, echoing Ruturaj, that XLA patch should go in the rocm/xla repo. Adding our own changes is the reason for having a branch there.

alekstheod:
> why do we need the xla patch? can't we create a pr in xla branch to fix?

Yes I did but for time being it is implemented as a patch to show the working version.

jax_rocm_plugin/build/build.py
alekstheod: I think release_wheel would be more appropriate here. no_rocm_rpath seems to be very specific. It is in fact replace rpaths with wheel based rpaths.

PR #301, opened 2026-02-10
Refactor build wheels pipeline to build wheels and run pytest

alekstheod:
This PR adjust the build_wheel pipeline to build wheel and make an integration test with installed wheel into the venv + using pytest.
Note: I tried - Use rbe platform from inside xla project, unfortunately it can't work as we have different tags in jax and xla doesn't handle these tags. Tags are used in xla to assign rbe pool to each test action!


Ruturaj4 (approved):

charleshofer (changes requested):

.github/workflows/ci.yml
```diff
@@ -15,95 +15,75 @@ concurrency:
...
-    uses: ./.github/workflows/build-wheels.yml
-    with:
-      python-versions: "3.11,3.12,3.13,3.14"
-      rocm-version: ${{ matrix.rocm-version }}
-      runner-label: '["linux-x86-64-1gpu-amd"]'
-    secrets:
-      rbe_ci_cert: ${{ secrets.RBE_CI_CERT }}
-      rbe_ci_key: ${{ secrets.RBE_CI_KEY }}
-      builder-image: "search"
-  call-build-docker:
```
charleshofer: Can you move this to the end of the workflow? We don't want to lose the ability for PR CI to check that the docker image build is busted

charleshofer (changes requested):

PR #303, opened 2026-02-11
Introduce asan build

alekstheod:
## Motivation
Introduce asan build for jax u-tests runs

## Technical Details
Added asan config, added execution mode ti ci-ut.yml workflow

## Test Plan
PR checks

### DIFF
--- a/jax_rocm_plugin/.bazelrc
+++ b/jax_rocm_plugin/.bazelrc
@@ -115,8 +115,8 @@ build:asan --linkopt="-L/usr/lib/llvm-18/lib/clang/18/lib/linux"
 build:asan --linkopt="-Wl,-rpath,/usr/lib/llvm-18/lib/clang/18/lib/linux"
 build:asan --linkopt="-lclang_rt.asan-x86_64"
 build:asan --linkopt="-lclang_rt.asan_cxx-x86_64"
-build:asan --@xla//build_tools/rocm:sanitizer=asan
-build:asan --run_under=//build/rocm:asan_wrapper
+build:asan --//build/rocm:sanitizer=asan
+build:asan --run_under=//build/rocm:sanitizer_wrapper
 build:asan --action_env TF_ROCM_AMDGPU_TARGETS="gfx908,gfx90a,gfx942"
 
 #############################################################################
--- a/jax_rocm_plugin/.bazelrc
+++ b/jax_rocm_plugin/.bazelrc
@@ -22,6 +22,8 @@ build --repo_env USE_HERMETIC_CC_TOOLCHAIN=1
 # TODO: Migrate for https://github.com/bazelbuild/bazel/issues/7260
 build:clang_local --noincompatible_enable_cc_toolchain_resolution
 build:clang_local --repo_env USE_HERMETIC_CC_TOOLCHAIN=0
+build:clang_local --linkopt="-fuse-ld=lld"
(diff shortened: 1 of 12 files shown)


jax_rocm_plugin/.bazelrc
```diff
@@ -99,6 +101,23 @@ build:rocm --copt=-Wno-gnu-offsetof-extensions
...
+build:asan --copt -fsanitize=address
+build:asan --copt -DADDRESS_SANITIZER
+build:asan --copt -O1
+build:asan --copt -g
+build:asan --copt -gsplit-dwarf
+build:asan --copt -fno-omit-frame-pointer
+build:asan --linkopt -fsanitize=address
+build:asan --linkopt -g1
+
+build:asan --linkopt="-L/usr/lib/llvm-18/lib/clang/18/lib/linux"
```
Arech8: Can you refactor a path to LLVM to a variable and use the var here and elsewhere instead? There are at least 3 uses of the path, it grows ugly...

Arech8 (changes requested):
I think, some generalization is very important here. At least to make a document with an explanation how to make a local build with all sanitizers. This shouldn't be an obscure knowledge that one must fish out of dozens of scripts and configs, but a clearly documented information instead.

jax_rocm_plugin/.bazelrc
alekstheod: Unfortunately bazelrc doesn't support declaring variables and reusing them. There are exactly 3 places where this substring is used. Not sure if it worth the effort to refactor. Additionally the real issue that we are still using systems compiler. Xla did switch to a hermetic llvm but we still do not support it! Once spport is there I believe we can get rid of these strings.

jax_rocm_plugin/.bazelrc
```diff
@@ -99,6 +101,23 @@ build:rocm --copt=-Wno-gnu-offsetof-extensions
...
+build:asan --linkopt="-Wl,-rpath,/usr/lib/llvm-18/lib/clang/18/lib/linux"
+build:asan --linkopt="-lclang_rt.asan-x86_64"
+build:asan --linkopt="-lclang_rt.asan_cxx-x86_64"
+build:asan --//build/rocm:sanitizer=asan
+build:asan --run_under=//build/rocm:sanitizer_wrapper
+build:asan --action_env TF_ROCM_AMDGPU_TARGETS="gfx908,gfx90a,gfx942"
```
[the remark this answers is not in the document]
alekstheod: If you wish to run on any other arch it is as simple as passing a new action_env in the end of your command line!
in bazel the last option wins so ```bazel test --config=rocm --config=asan --repo_env=TF_ROCM_AMDGPU_TARGETS=gfx945``` will override what is in a config, this is well documented in bazel itself!

The only thing you need to do is ```bazel test --config=rocm --config=asan``` (Do we really need a document for that?) The rest is in bazel docu: 
image


https://bazel.build/reference/command-line-reference#options-specifying-or-altering-a-generic-input-to-a-bazel-command

The script itself is just a bazel command there is no custom code in it to document something. Configs are also standard bazel configs, well documented inside the bazel itself.

Arech8 (approved):

Arech8 (changes requested):
please fix the linter first

Arech8 (approved):

PR #330, opened 2026-03-02
Fix symlink creation for newer builds of TheRock

charleshofer:
Earlier versions of ROCm 7 were missing a link to amdgcn, so we created the link when we installed ROCm via the `get_rocm.py` script. However, on newer builds of TheRock, the link is there. If amdgcn already exists, the link creation fails and causes the build script to fail when it doesn't need to. Just wrap the link creation in a `try` block so the ROCm install continues.

rocm-jaxlib-v0.8.0 is an old branch whose wheels have already been released, but TheRock continues to build from it for testing.


JeniferC99 (approved):
Tested locally, it works

PR #337, opened 2026-03-03
Update TE to new build

gabeweisz:
## Motivation

Updates TransformerEngine to a new build
Also builds both GFX950 and GFX942 architectures for TE during CI so we can use these artifacts for both platforms


psanal35 (approved):
Looks good to me. Thanks~

PR #352, opened 2026-03-17
Shrink manylinux image

charleshofer:
Make the manylinux image a little smaller by only installing the components of LLVM that we need (clang and lld)


mminutoli (commented):
Premise: this is fine.

I wonder if there is an easier way with `llvm.sh` from the instructions here https://apt.llvm.org/

PR #368, opened 2026-03-28
Rebuild ROCR-Runtime with intercept queue fix in Docker base image

magaonka-amd:
- Add docker/patches/rocr-intercept-queue-fix.patch with the staging buffer and std::mutex fix from ROCm/rocm-systems#2185
- Add a RUN block to Dockerfile.base-ubu24 that sparse-clones rocm-systems at the matching rocm-${ROCM_VERSION} tag, applies the patch, builds libhsa-runtime64.so with ROCm LLVM clang, and installs it over the packaged version
- The profiler SDK intercept queue corrupts AQL packets when they wrap around the ring buffer boundary, causing HSA_STATUS_ERROR_INVALID_PACKET_FORMAT aborts under parallel GPU workloads (e.g. pytest-xdist with 16 workers on 8 GPUs)
- Remove this block when the fix ships in a ROCm release

### DIFF
--- a/docker/Dockerfile.base-ubu24
+++ b/docker/Dockerfile.base-ubu24
@@ -148,6 +148,43 @@ RUN [ "$ROCM_VERSION" = "7.1.1" ] || (       \
 # This mitigates crashes related to the kernel database.
 ENV MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE
 
+# Temporarily include a custom built ROCR-Runtime. This should be removed
+# as soon as https://github.com/ROCm/rocm-systems/pull/2185 lands
+# in a ROCm release that we can install as deb package or with therock.
+# The profiler SDK's intercept queue corrupts AQL packets when they wrap
+# around the ring buffer boundary, causing HSA_STATUS_ERROR_INVALID_PACKET_FORMAT
+# aborts under parallel GPU workloads.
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.base-ubu24
```diff
@@ -148,6 +148,35 @@ RUN [ "$ROCM_VERSION" = "7.1.1" ] || (       \
...
+    git apply /tmp/rocr-fix.patch &&                   \
+    cd projects/rocr-runtime && mkdir build && cd build && \
+    cmake -DCMAKE_INSTALL_PREFIX=${ROCM_PATH}          \
+          -DCMAKE_PREFIX_PATH=${ROCM_PATH}             \
+          -DCMAKE_BUILD_TYPE=Release                   \
+          -DBUILD_SHARED_LIBS=ON                       \
+          -DCMAKE_C_COMPILER=${ROCM_PATH}/llvm/bin/clang \
+          -DCMAKE_CXX_COMPILER=${ROCM_PATH}/llvm/bin/clang++ .. && \
+    make -j$(nproc) &&                                 \
+    cp rocr/lib/libhsa-runtime64.so.* ${ROCM_PATH}/lib/ && \
```
mminutoli: Maybe overkill, but I'd follow the same approach for miopen in L136-L138 using dpkg-divert as dpkg does not like human touching files it thinks to own.

PR #379, opened 2026-04-08
Adjust unversioned Python to symlink to Python 3.11.

tsrw2048:
## Motivation
Originally, symlinking Python to Python3 resulted in a symlink to Python 3.6. In order to use a more recent version, we specify Python 3.11. This will satisfy some of the version dependencies needed in the JAX CI scripts.

## Changes
The only change is to file: `docker/manylinux/Dockerfile.jax-manylinux_2_28-rocm`. The version of Python 3 that is being symlinked is specified as Python 3.11 which was already on the manylinux image.


magaonka-amd (approved):
LGTM but keep an eye on next continuous CI.

PR #380, opened 2026-04-08
Bootstrap pip and install auditwheel on manylinux image

tsrw2048:
## Motivation
JAX CI scripts require the use of auditwheel so we bootstrap pip and use it to install auditwheel on the manylinux image.

## Changes
The change is to file: `docker/manylinux/Dockerfile.jax-manylinux_2_28-rocm`. We add a line setting up pip and use it to install auditwheel.


psanal35 (approved):
LGTM~

PR #388, opened 2026-04-16
Switch jax 0.8.2 build to 0.9.1 for Llama perf test

gabeweisz:
Switch jax 0.8.2 build to 0.9.1 for Llama perf test per customer request

### DIFF
--- a/.github/workflows/llama-perf.yml
+++ b/.github/workflows/llama-perf.yml
@@ -128,7 +128,6 @@ jobs:
                 jaxlib==0.6.0 \
                 "$PJRT_URL" \
                 "$PLUGIN_URL"
-            fi
             elif [ "${{ matrix.jax-version }}" == "0.9.1" ]; then
               apt update
               apt install libdw1 libglib2.0-0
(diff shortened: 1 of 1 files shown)


.github/workflows/llama-perf.yml
```diff
@@ -124,6 +128,10 @@ jobs:
                 jaxlib==0.6.0 \
                 "$PJRT_URL" \
                 "$PLUGIN_URL"
+            elif [ "${{ matrix.jax-version }}" == "0.9.1" ]; then
+              apt update
+              apt install libdw1 libglib2.0-0
+              pip install --force-reinstall jax==0.9.1 jax-rocm7-plugin jax-rocm7-pjrt
```
gulsumgudukbay: jax-rocm7-plugin and jax-rocm7-pjrt should be =~0.9.1 to grab the latest post release

gulsumgudukbay (changes requested):

.github/workflows/llama-perf.yml
gabeweisz: Done - from what [I can tell](https://pip.pypa.io/en/stable/reference/requirement-specifiers/#examples) it needs to be ~= and not =~

gulsumgudukbay (approved):
looks good thanks!

PR #391, opened 2026-04-17
RFC: S3 artifact layout for multi-version ROCm CI

mminutoli:
## Summary

This is an RFC (empty commit, no code changes) to discuss the S3 artifact layout needed to support building and testing across multiple ROCm versions (deb-based releases and TheRock pip wheels).

Pipeline 1 (`test-and-upload.yml` with GitHub Actions artifacts + direct MySQL insert) is being retired. We need a clean, unified S3 layout that groups both **wheel artifacts** and **test logs** by ROCm version.

---

## Problem: Wheel Name Collisions

`get_rocm_version_flag()` in `jax_rocm_plugin/build/rocm/tools/build_wheels.py` reduces all 7.x versions to `"7"`:

```python
def get_rocm_version_flag(rocm_version):
    version_string = rocm_version[0]
    if version_string == "6":
        version_string = "60"
    return version_string
```

All ROCm 7.x builds produce identically-named wheels:
- `jax_rocm7_plugin-*.whl`
- `jax_rocm7_pjrt-*.whl`
- `jaxlib-*+rocm7-*.whl`

Today, GitHub Actions artifact names include the full version (`plugin_wheels_r7.1.1`), preventing collision. The S3 layout must also enforce this separation.

---

### Wheels (`build_rocm_artifacts.yml` in `ROCm/jax`)

```
s3://jax-ci-amd/jax-rocm-plugin-wheels/    (flat, no version separation)
    jax_rocm7_plugin-*.whl
    jax_rocm7_pjrt-*.whl
s3://jax-ci-amd/rocm-wheels/LATEST          (single global pointer)
```

- **Collision risk**: if builds for ROCm 7.1.1 and 7.2.0 both upload, the later one silently overwrites.
- **`LATEST` pointer**: global, not per-version.
- Source: `jax/.github/workflows/build_rocm_artifacts.yml`

### Test logs (`upload_rocm_logs.sh` in `ROCm/jax`)

```
s3://jax-ci-amd/jax-ci-test-logs/
    {org/repo}/{branch}/{nightly|continuous}/{DATE}_{run_id}_{attempt}/{combo}/
        run-manifest.json
        logs.tar.gz
        _SUCCESS
```

- `combo` = `py{X_Y}-rocm{X_Y_Z}-gpu_{N}` (ROCm version is in the leaf, not a grouping tier)
- Browsing all results for a specific ROCm version requires listing every `run_key` and parsing combo names.
- Source: `jax/ci/upload_rocm_logs.sh`

### Test log ingestion (`rocm-jax`)

- `ci/ingest_jax_ci_logs.sh` scans `jax-ci-test-logs/{REPO}/` recursively for `_SUCCESS` markers
- `ci/upload_pytest_to_db.py` extracts `(run_key, combo)` as the last two path segments from `artifact_uri`
- Dedup key: `(github_repository, github_ref_name, is_nightly, run_key, combo)` -- **no `rocm_version`**
- Source: `ci/ingest_jax_ci_logs.sh`, `ci/upload_pytest_to_db.py`

### Wheel download (`ROCm/jax`)

- `download-jax-rocm-wheels` action resolves `LATEST` via CloudFront, then greps for `jax_rocm{major}_*`
- No ROCm minor version awareness
- Source: `jax/.github/actions/download-jax-rocm-wheels/action.yml`

---

## Proposed Unified Layout

```
s3://jax-ci-amd/{REPO}/{branch}/{nightly|continuous}/rocm-{version}/{run_key}/
    wheels/
        jax_rocm7_plugin-*.whl
        jax_rocm7_pjrt-*.whl
        jaxlib-*+rocm7-*.whl
    tests/{combo}/
        run-manifest.json
        logs.tar.gz
        _SUCCESS
        _INGESTED

s3://jax-ci-amd/rocm-wheels/rocm-{version}/LATEST    (per-version pointer)
```

### Design Points

- **`rocm-{version}`** as a first-class path tier prevents wheel collisions and enables per-version browsing/cleanup (e.g., `rocm-7.1.1`, `rocm-7.2.0`, `therock-7.13.0a20260401`)
- **`run_key`** stays `{DATE}_{run_id}_{attempt}` for time-based retention
- **`wheels/` and `tests/{combo}/`** under the same `run_key` keeps build+test artifacts together
- **`LATEST` pointer becomes per-version** so consumers resolve the latest wheels for a specific ROCm version
- Manifest remains source of truth for detailed metadata
- `_SUCCESS` / `_INGESTED` marker pattern preserved

---

### Upstream `ROCm/jax` (3 items)

| # | Component | Change |
|---|-----------|--------|
| 1 | `build_rocm_artifacts.yml` | Version-aware `s3_upload_uri`; per-version `LATEST` pointer; parameterize container image (currently hardcoded to `rocm-7.2.0`) |
| 2 | `upload_rocm_logs.sh` | Add `rocm-{version}/` tier to `PREFIX` between `{nightly\|continuous}` and `{run_key}` |
| 3 | `download-jax-rocm-wheels/action.yml` | Resolve `LATEST` from `rocm-wheels/rocm-{version}/LATEST` instead of global pointer |

### `rocm-jax` (5 items)

| # | Component | Change |
|---|-----------|--------|
| 4 | `ci/ingest_jax_ci_logs.sh` | Adjust `RUN_DIR`/`RUN_DATE` extraction for new path depth; add optional `FILTER_ROCM_VERSION`; support both old+new paths during transition |
| 5 | `ci/upload_pytest_to_db.py` | Update `parse_run_key_and_combo()` for new path structure; add `rocm_version` to `find_existing_run_id()` dedup query |
| 6 | DB schema (`jax_ci_runs`) | Add `rocm_version` to uniqueness constraint (if not there already) |
| 7 | `pytest-results-to-db.yml` | Add optional `rocm-version` input; parameterize `INPUT_GPU_ARCH` (currently hardcoded to `MI350`) |
| 8 | Pipeline 1 cleanup | Remove `ci/upload_test_to_db.py`, `test-and-upload.yml`, and related `ci_runs`/`ci_tests`/`ci_results` tables |

---

## Migration

Existing test logs are self-describing via `run-manifest.json`. A migration script can:
1. List all `_SUCCESS` markers under old prefix
2. Read `rocm_version` from each `run-manifest.json`
3. `aws s3 mv` each prefix to the new path structure

Risks:
- **`artifact_uri` in DB** will point to old paths for historical rows
- **Race during migration** if upstream is still writing old-format paths -- mitigate by having ingestion scan both old and new prefixes during transition

---

## Open Questions

1. Should TheRock nightlies use `therock-{nightly_version}` or a different naming convention for the version tier?
2. Retention policy per version tier (keep last N nightlies? keep all releases?)
3. Should wheel upload land directly in the unified layout, or go through a staging prefix first?
4. How long should we support both old and new S3 paths in the ingestion script?
5. Should the `COMBO` segment drop the `rocm{X_Y_Z}` part now that ROCm version is in the path?
6. How do we make this or can we make this our release process?


psanal35:
Just to confirm: are we expecting to support multiple ROCm versions in parallel in CI (e.g. different 7.x releases and/or TheRock builds)? If so, could we also clarify the ROCm version identifier/format?

psanal35:
> Should the `COMBO` segment drop the `rocm{X_Y_Z}` part now that ROCm version is in the path?

If yes, we should also update `pytest_rocm.yml` accordingly, since the `COMBO` string is constructed there.

psanal35:
I agree with the collision risk around the `LATEST` pointer and wheels folder. Including an explicit `{org}/{repo}/{branch}` prefix would further help avoid cross‑repo wheel collisions.

mminutoli:
> I agree with the collision risk around the `LATEST` pointer and wheels folder. Including an explicit `{org}/{repo}/{branch}` prefix would further help avoid cross‑repo wheel collisions.

With the premise that I am no expert on cloud storage, my understanding is that deep hierarchies are to be avoided. Nevertheless, I like your suggestions. It would prevent that our downstream repo messes up upstream CI.

Regarding the COMBO, yes it could be simplified in the case we will decide to move the rocm version as part of the path hierarchy.

@charleshofer @gulsumgudukbay : what would be required for, or what would simplify, our release process?

charleshofer:
> Should TheRock nightlies use therock-{nightly_version} or a different naming convention for the version tier?

I think the nightly version is fine

> Retention policy per version tier (keep last N nightlies? keep all releases?)

30 days for everything except releases, which get kept around all the time. Not sure that we should even keep release wheels in this bucket.

> Should wheel upload land directly in the unified layout, or go through a staging prefix first?

I'd just land it in the unified layout right away. Keep it simple.

> How long should we support both old and new S3 paths in the ingestion script?

An `if` statement that checks for both the new and old S3 path? I don't see why we need to support both. For now, we can upload artifacts to both S3 paths, change the ingestion script to use the new path, and then remove the artifact upload for the old S3 path.

> Should the COMBO segment drop the rocm{X_Y_Z} part now that ROCm version is in the path?

Could you show an example of what the new path would look like with the COMBO? It's not clear to me from the write-up

> How do we make this or can we make this our release process?

I think it's fine to keep release artifacts here if they then get promoted from here to either PyPI or an official AMD Python package index. I think it's still not clear what that package index is supposed to be. There's the Radeon index that was part of the legacy ROCm build pipeline, a potential training "mono-repo" build process, and TheRock's package index.

charleshofer:
> what would be required for, or what would simplify, our release process?

The best thing I think we can do for our release process is to settle on just one place to release wheels. It matters less whether that's the Radeon repo, the training mono-repo index, TheRock's index, our own S3 index, or PyPI. It's kind of all over the place and confusing for users. Unless we've got buy-in from TheRock, ROCm DevOps, and Jason, I'm reluctant to mark anything in our S3 as a "release" wheel and create yet another release channel.

mminutoli:
> > Should the COMBO segment drop the rocm{X_Y_Z} part now that ROCm version is in the path?
> 
> Could you show an example of what the new path would look like with the COMBO? It's not clear to me from the write-up

It's about the way we store logs. the COMBO section of the path capture the point of the CI matrix (e.g., py3_12-rocm7_2_0-gpu_4).

Here is the line where it gets set: https://github.com/jax-ml/jax/blob/5ff6cfdd78e00db8b37a732f8c71008d8bf983c1/ci/upload_rocm_logs.sh#L60

charleshofer:
> > > Should the COMBO segment drop the rocm{X_Y_Z} part now that ROCm version is in the path?
> > 
> > 
> > Could you show an example of what the new path would look like with the COMBO? It's not clear to me from the write-up
> 
> It's about the way we store logs. the COMBO section of the path capture the point of the CI matrix (e.g., py3_12-rocm7_2_0-gpu_4).
> 
> Here is the line where it gets set: https://github.com/jax-ml/jax/blob/5ff6cfdd78e00db8b37a732f8c71008d8bf983c1/ci/upload_rocm_logs.sh#L60

Yeah, then I think we should drop it. The ROCm version should only appear in one place in the path.

PR #392, opened 2026-04-20
fixwheel: also handle manylinux_2_27 platform tag

WBobby:
## Motivation

Add "manylinux_2_27" to the platform tag check so these wheels are
also stripped to linux_x86_64 before auditwheel repairs them to
manylinux_2_28.


gulsumgudukbay (approved):
LGTM.
Note: 
If upstream ever bumps build_utils.py:115 to manylinux_2_31 (or similar), this fix needs updating again. A safer formulation:
`if "manylinux" in plat_tag and "manylinux_2_28" not in plat_tag:`

PR #397, opened 2026-04-20
Support ingestion of multiple pytest JSON reports for a single CI run

psanal35:
This PR supports ingestion of multiple pytest JSON reports for **each CI run**. Upstream may now split tests into single- and multi-GPU runs, producing two reports under the same logs directory. The previous logic assumed exactly one report.

It now:
-Accepts multiple pytest JSON reports per run
-Aggregates test results into one run before loading
-Uses the latest report timestamp as run-metadata
-Preserves existing behavior for 0 or 1 pytest-report

Note that there is:
-No DB schema changes
-No changes to existing ingestion flow
-**No overlap** in test `nodeids` across reports

### DIFF
--- a/ci/upload_pytest_to_db.py
+++ b/ci/upload_pytest_to_db.py
@@ -2,7 +2,7 @@
 """
 Ingest JAX Pytest results from S3 into MySQL.
 
-Recursively locates one Pytest report under given log dir
+Recursively locates up to two Pytest report under given log dir
 
 Tables:
  - jax_ci_runs:    one row per run
@@ -102,8 +102,12 @@ def parse_run_key_and_combo(artifact_uri: str) -> tuple[str, str]:
 # -----------------------------
 # Input/File Loading
 # -----------------------------
-def find_pytest_report_json(local_logs_dir: Path) -> Optional[Path]:
-    """Find the single Pytest JSON report under the given logs directory, if present"""
+def find_pytest_report_jsons(local_logs_dir: Path) -> List[Path]:
+    """Find up to two Pytest JSON reports under the given logs directory.
+
+    A run may produce one report (single-GPU) or two reports (single + multi-GPU).
+    Returns a sorted list of paths; empty if no reports are found.
+    """
     ignore = {MANIFEST_FILENAME, "last_running.json"}
     reports = [
         p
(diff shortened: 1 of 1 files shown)


ci/upload_pytest_to_db.py
```diff
@@ -113,15 +117,15 @@ def find_pytest_report_json(local_logs_dir: Path) -> Optional[Path]:
         and not p.name.endswith("last_running.json")
     ]
     if not reports:
-        return None
-    if len(reports) != 1:
+        return []
+    if len(reports) > 2:
```
magaonka-amd: Pakize, thanks for the fix.
your code is already capable of aggregating more than two result JSONs.
so it is easy to extend.

I suggest you dont cap it at two here.

my suggestion is to to have a variable like this MAX_REPORTS_PER_RUN = 2

and then do if len(reports) > MAX_REPORTS_PER_RUN

This way in future if we add another pass ( which can be very much possible case )
All you need to do is change  one variable MAX_REPORTS_PER_RUN.

What do you think?

ci/upload_pytest_to_db.py
```diff
@@ -636,12 +640,16 @@ def upload_pytest_results(  # pylint: disable=too-many-locals
...
+    reports = find_pytest_report_jsons(local_logs_dir)
+    report_created_at = None
+    tests = []
+
+    for report in reports:
+        current_created_at, current_tests = load_from_pytest_json(report)
+        if current_created_at is not None:
+            if report_created_at is None or current_created_at > report_created_at:
+                report_created_at = current_created_at
+        tests.extend(current_tests)
```
magaonka-amd, line 654: does this make sure no overlap in nodeids?

ci/upload_pytest_to_db.py
psanal35: done.

ci/upload_pytest_to_db.py
psanal35: With the bash script change, the tests are explicitly split into separate single- and multi-GPU passes. Given that split, nodeids should not overlap across the generated reports.

magaonka-amd (approved):
LGTM thanks

PR #398, opened 2026-04-20
Build: clean manylinux_2_28 wheels + https for repo.radeon.com

gulsumgudukbay:
## Motivation

Three build/CI fixes:

1. Building the new builder image takes ages and sometimes errors out from timeout due to http mirrors being extremely slow. Switch http to https for `repo.radeon.com`.
2. Builds with `--wheel-post-release` crash mid-fixwheel after PR #392 (`No hash found for file '<...>/WHEEL'`).
3. On the runs where they do complete, `wheelhouse/` ends up with three differently- tagged copies of the same package (`…-linux_x86_64.whl`, `…-manylinux_2_27_x86_64.whl`, `…-manylinux_2_28_x86_64.whl`). TheRock's release-promotion filter only matches the clean `…-manylinux_2_28_x86_64.whl`, so the noisy output silently breaks the release path.

## Technical Details

- `tools/get_rocm.py`: http changed to https on line 430.
- `_apply_wheel_post_release` in `jax_rocm_plugin/pjrt/tools/build_utils.py` and `jax_rocm_plugin/jaxlib_ext/tools/build_utils.py` now emits PEP 376-compliant RECORD with `path,sha256=<urlsafe-b64>,size` entries; previously it wrote `path,,` for every entry, which made `wheel tags` (called by `fixwheel.py`) abort.
- `fix_wheel` in `jax_rocm_plugin/build/rocm/tools/fixwheel.py` now removes the Bazel input (`…-manylinux_2_27_x86_64.whl`) and the `wheel tags` intermediate (`…-linux_x86_64.whl`) from `dist_manylinux/` after `auditwheel repair` succeeds, so `build_wheels.py`'s blind `shutil.copy(dist_manylinux/*, wheelhouse/)` step doesn't propagate them alongside the audited output.

## Test Plan and Result

Ran ci_build, which exercises both the builder Docker rebuild and the wheel pipeline:

## Submission Checklist

### DIFF
--- a/jax_rocm_plugin/build/rocm/tools/fixwheel.py
+++ b/jax_rocm_plugin/build/rocm/tools/fixwheel.py
@@ -61,6 +61,8 @@ def parse_wheel_name(path):
 
 def fix_wheel(path):
     """Fixes a wheel and attaches manylinux platform labels to it"""
+    original_input = path
+    intermediate = None
     tup = parse_wheel_name(path)
     plat_tag = tup[4]
     if "manylinux2014" in plat_tag or "manylinux_2_27" in plat_tag:
@@ -80,6 +82,7 @@ def fix_wheel(path):
         new_wheel = output.stdout.strip()
         new_path = os.path.join(os.path.dirname(path), new_wheel)
         LOG.info("Stripped broken tags and created new wheel at %r", new_path)
+        intermediate = new_path
         path = new_path
 
     # build excludes, using auditwheels lddtree to find them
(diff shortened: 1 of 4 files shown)


jax_rocm_plugin/build/rocm/tools/fixwheel.py
```diff
@@ -104,6 +107,23 @@ def fix_wheel(path):
...
+    # input and any `wheel tags` intermediate so build_wheels.py doesn't
+    # propagate them alongside the auditwheel output (which would leave
+    # users with three differently-tagged copies of the same package in the
+    # wheelhouse).
+    if intermediate and intermediate != original_input:
+        try:
+            os.remove(intermediate)
+            LOG.info("Removed intermediate wheel %r", intermediate)
+        except OSError as exc:
+            LOG.warning("Could not remove intermediate %r: %s", intermediate, exc)
```
magaonka-amd, line 120: should this be fatal or just warning?
other than that LGTM

gulsumgudukbay: > should this be fatal or just warning? other than that LGTM

I dont think it should be fatal, we are just removing intermediate wheels afterall

magaonka-amd (approved):

PR #399, opened 2026-04-21
Upgrade ROCm base images from 7.2.0 to 7.2.2

Ruturaj4:
Migrate the ubu24 base/dev images and the manylinux_2_28 builder image from ROCm 7.2.0 to 7.2.2 in the build-base-docker.yml matrices.

Skip the custom MIOpen and ROCR-Runtime rebuild blocks for 7.2.2 in Dockerfile.base-ubu24 since the upstream fixes they work around have already landed:
- MIOpen stack-overflow kernel init: ROCm/rocm-libraries#4472 (merged into develop on 2026-02-20)
- ROCR intercept-queue AQL packet fix: ROCm/rocm-systems#2185 (merged into develop on 2025-12-05)

Both were merged before 7.2.2 was cut, so ROCm 7.2.2 ships the fixed libMIOpen.so and libhsa-runtime64.so directly from the package repo.

## Submission Checklist


docker/Dockerfile.base-ubu24
```diff
@@ -129,7 +129,8 @@ RUN --mount=type=cache,target=/var/cache/apt   \
 # Temporarily include a custom built MIOpen. This should be removed
 # as soon as https://github.com/ROCm/rocm-libraries/pull/4472 lands
 # in a ROCm release that we can install as deb package or with therock.
-RUN [ "$ROCM_VERSION" = "7.1.1" ] || (       \
+# Fix is already included in ROCm 7.2.2 and later, so skip the rebuild there.
+RUN [ "$ROCM_VERSION" = "7.1.1" ] || [ "$ROCM_VERSION" = "7.2.2" ] || (       \
```
charleshofer, line 133: We're changing the workflow to only build for 7.1.1 and 7.2.2, so can we just remove this big `RUN` command that does the MIOPen build?

jax_rocm_plugin/platform/linux/BUILD
```diff
@@ -28,18 +28,3 @@ platform(
         "Pool": "linux_x64_gpu",
     },
 )
```
charleshofer, line 30: It doesn't matter that much since we don't build from this repo anymore, but why are we removing a bunch of RBE stuff? It seems unrelated to switching to ROCm 7.2.2 and out of scope for the PR. Was the PR CI failing when trying to run tests?

tools/get_rocm.py
```diff
@@ -372,6 +372,38 @@ def _build_installer_url(rocm_version, metadata):
...
+
+    probe_url = "https://repo.radeon.com/graphics/%s/" % rocm_version_str
+    try:
+        req = urllib.request.Request(probe_url, method="HEAD")
+        with urllib.request.urlopen(req, timeout=10) as response:
+            if 200 <= response.status < 400:
+                LOG.info("Graphics repo: using full-version path %s", rocm_version_str)
+                return rocm_version_str
+    # pylint: disable=W0718
+    except Exception as exc:
```
charleshofer: nit: Could we catch an `HTTPError` here instead of the very general `Exception`?

charleshofer (changes requested):

docker/Dockerfile.base-ubu24
Ruturaj4: @charleshofer not only that, we do have fixes for the rocprof sdk

jax_rocm_plugin/platform/linux/BUILD
Ruturaj4: becuase it fails in CI. and this we should remove this stuff anyways as it doesn't reflect the env that we are using to run the upstream CI.

charleshofer (approved):

psanal35:
No need to delete the workflow files; they are already disabled.

magaonka-amd (approved):
Looks good to me.

PR #415, opened 2026-05-01
Release work for rocm-jaxlib-v0.9.2

magaonka-amd:
-This patch prepares rocm-jax for 9.2 release
-upgrades bazel version
-fixes python_rules issue due to recent rule upgrade in XLA
-updates SHA to point to latest XLA and JAX hashes


.github/workflows/llama-perf.yml
```diff
@@ -42,17 +42,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jax-version: ["0.6.0", "0.9.1"]
+        jax-version: ["0.6.0", "0.9.2"]
```
mminutoli, line 45: why are we keeping 0.6.0? is this THE baseline?

mminutoli (approved):
LGTM.

my only worry is that things might get out of sync with upstream.

gulsumgudukbay (approved):
LGTM but waiting the CI issues to resolve

PR #426, opened 2026-05-07
fix(docker): scope MIOpen patch to ROCm 7.2.x only

mminutoli:
## Summary
- Restricts the custom MIOpen from-source rebuild to only the ROCm 7.2.x series, rather than running for every version except 7.1.1.
- Uses POSIX shell parameter expansion (`${ROCM_VERSION%.*}`) to check major.minor, so 7.2.0, 7.2.1, etc. trigger the patch while 7.1.x, 7.12.x, and beyond skip it.

## Test plan
- [x] Build the base image with `ROCM_VERSION=7.2.0` — MIOpen patch should execute.
- [x] Build the base image with `ROCM_VERSION=7.1.1` — MIOpen patch should be skipped.


mminutoli:
Here is the CI run for the rock: https://github.com/ROCm/TheRock/actions/runs/25525140576/job/74918827575

Ruturaj4 (approved):

psanal35 (approved):
LGTM~

PR #433, opened 2026-05-12
Add warning to README

charleshofer:
We've had a request to add a matrix to let people know which repo they should build JAX from. Adding that here to the README.


gulsumgudukbay (approved):
Looks nice! thanks

PR #435, opened 2026-05-13
Fix ROCm version formatting in build container env vars

mdnaveedh:
Follow-up to #430. The '+' -> '.' sanitization PR #430 added for the docker image tag is also needed for the ROCM_VERSION_EXTRA / ROCM_VERSION env vars passed to the build container, which setup.py concatenates into the wheel version. 

Without it, TheRock builds whose
ROCM_VERSION_EXTRA contains a '+' (e.g. "7.13.0.dev0+<gitsha>") produce an invalid PEP 440 version like "0.9.1+rocm7.13.0.dev0+<sha>" and crash setuptools with packaging.version.InvalidVersion.

### DIFF
--- a/jax_rocm_plugin/build/rocm/ci_build
+++ b/jax_rocm_plugin/build/rocm/ci_build
@@ -114,6 +114,11 @@ def dist_wheels(
     # NOTE: GIT_DIR and GIT_WORK_TREE are NOT set because they interfere with
     # Bazel's git_repository rule. Instead, we pass ROCM_JAX_COMMIT explicitly.
     rocm_jax_commit = get_rocm_jax_commit()
+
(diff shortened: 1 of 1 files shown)


gulsumgudukbay (changes requested):
Hi @mdnaveedh  could you please add why this change is needed? I want to see the error log and also test method and test result. 
Also why is the version containing `dev`?

gulsumgudukbay (approved):
Verified with dist_wheels using --rocm-version=7.13.0.dev0+abc123. Base rocm-jaxlib-v0.9.1 fails with packaging.version.InvalidVersion from 0.9.1+rocm7.13.0.dev0+abc123; this PR sanitizes the env vars used by setup.py and the build succeeds. This should be a no-op for normal ROCm versions without +.

PR #444, opened 2026-05-20
Remove sysdeps entries from rpath and Update XLA hash

gulsumgudukbay:
## Motivation

PR #438 removed /opt/rocm/lib/rocm_sysdeps/lib from the wheel RUNPATH, breaking TheRock CI where ROCm is installed via tarball extraction to /opt/rocm. The librocm_sysdeps_*.so.1 libraries live at /opt/rocm/lib/rocm_sysdeps/lib/ in that layout and could no longer be found.

## Technical Details

Added /opt/rocm/lib/rocm_sysdeps/lib back to the RUNPATH in both build_gpu_kernels_wheel.py and build_gpu_plugin_wheel.py, alongside the existing /opt/rocm/lib entry. This covers TheRock tarball installs (extracted to /opt/rocm-<ver>, symlinked to /opt/rocm) and the three ROCm install methods: pip wheels, tarballs, and system packages (deb/rpm).

Fixes https://github.com/ROCm/TheRock/issues/5369.

## Test Plan

Verify ldd resolves all deps when ROCm is installed via tarball at /opt/rocm.

## Test Result

readelf -d confirms /opt/rocm/lib/rocm_sysdeps/lib is present in the patched .so RUNPATH.

## Submission Checklist

### DIFF
--- a/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
+++ b/jax_rocm_plugin/jaxlib_ext/tools/build_gpu_kernels_wheel.py
@@ -257,18 +257,16 @@ def prepare_wheel_rocm(wheel_sources_path: pathlib.Path, *, cpu, rocm_version, s
         f"_triton.{pyext}",
         f"rocm_plugin_extension.{pyext}",
     ]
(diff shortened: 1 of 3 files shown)


magaonka-amd (approved):
This is literal dance !!! Change looks good

PR #449, opened 2026-05-21
[TheRock] Align manylinux Dockerfile with ROCm counterpart

mminutoli:
## Summary

- Copy `clang.cfg` into `/opt/rocm/llvm/bin/` for both `clang` and `clang++`, mirroring what the ROCm Dockerfile already does for `/usr/lib/llvm-18/bin/`
- Add a symlink `libstdc++.so.6` → `libstdc++.so` so the hermetic `ld.lld` linker can resolve `-lstdc++` on AlmaLinux 8

Note: the TheRock Dockerfile already sets up an unversioned `python` symlink (pointing to `/opt/python/cp312-cp312/bin/python3`) and installs `auditwheel` earlier in the file, so those steps were not ported from the ROCm counterpart.

## Test plan

- [x] Verify the TheRock manylinux Docker image builds successfully
- [x] Verify JAX CI scripts run correctly inside the built image


psanal35 (approved):
Looks good to me~

magaonka-amd (approved):
LGTM. you can verify presubmit before merge

PR #456, opened 2026-06-03
[CI] Add image-set input to selectively build/repush base images

gulsumgudukbay:
## Motivation

The workflow builds and pushes every image set in one run - the legacy apt-ROCm (rocm720) images and the TheRock images. The rocm720 images are shared with jax-ml/jax upstream (it pins rocm-tag: rocm720), so there's no way to refresh just the TheRock images without also repushing the upstream-shared ones. This adds per-set selection for manual runs.

## Technical Details

- Adds a workflow_dispatch input image-set (therock / legacy / both, default both).
- Gates each job: legacy jobs (build-base-images, build-manylinux-builder-images) run for legacy/both; TheRock jobs (build-therock-base-images, build-therock-manylinux-images) run for therock/both.
- Only narrows manual workflow_dispatch runs; scheduled/push/pull_request still build every set. Composes with push-latest-tag.

## Test Plan

- YAML parses; all four jobs carry the expected gate.
- On dispatch: therock runs only TheRock jobs, legacy only legacy jobs, both runs all four; non-dispatch runs build all.

## Test Result

Local YAML validation passed; gating confirmed on all four jobs. Dispatch include/skip behavior to confirm on first manual run post-merge.

## Submission Checklist

### DIFF
--- a/.github/workflows/build-base-docker.yml
+++ b/.github/workflows/build-base-docker.yml
@@ -11,6 +11,17 @@ on:
         required: false
         type: boolean
         default: false
+      image-set:
+        description: >-
+          Which image set to build/repush on manual dispatch: 'legacy'
+          (rocm720), 'therock', or 'both'. Scheduled/push/PR build all.
+        required: false
+        type: choice
+        options:
+          - therock
+          - legacy
+          - both
+        default: both
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


mminutoli (approved):
LGTM

PR #457, opened 2026-06-04
Add hermetic llvm docker images for JAX/XLA for ROCm and theRock

hsharsha:
Add unified docker images for CI builds


docker/Dockerfile.theRock
```diff
@@ -0,0 +1,70 @@
...
+# conventional /opt/rocm symlink so ENV directives use a known path.
+RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
+    python3 -m pip install --break-system-packages \
+        --pre --index-url "${THEROCK_INDEX_URL}" \
+        "rocm[libraries,devel]${THEROCK_VERSION:+==}${THEROCK_VERSION}" && \
+    rocm-sdk init && \
+    ln -sfn "$(rocm-sdk path --root)" /opt/rocm
+
+# Fix broken link to ld.lld in theRock
+RUN ln -sf /opt/rocm/llvm/bin/lld /opt/rocm/llvm/bin/ld.lld
```
i-chaochen, line 91: I'm sorry I still couldn't fully got it why there is a broken link to `ld.lld` in theRock. 

XLA uses inprocess lld https://github.com/openxla/xla/blob/main/xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc#L619-L634 

is any path changes on theRock? or do we need to modify there for theRock and add fallback for legacy ROCm? cc @draganmladjenovic

i-chaochen: because theRock is released by many arch and it uses `lld` while xla only rely on `ld.lld` ?

hsharsha: @i-chaochen  On rocm there is a soft link from /opt/rocm/llvm/bin/ld.lld -> /opt/rocm/llvm/bin/lld present which is not there in theRock installation. As a result 4 of the XLA tests failed at runtime as it was unable to use linker. Above fixes that issue in theRock. Currently we are using singe arch gfx950 theRock as there are issue with multi arch whls.

i-chaochen: Thanks for the clarification! Since rocm can have this soft link, why theRock doesn't have it anymore? do we need to add to somewhere in [theRock](https://github.com/ROCm/TheRock/tree/main)?

docker/python_requirements.txt
```diff
@@ -0,0 +1,18 @@
...
+
+# For XLA
+pyyaml
+
+# For JAX
+build ~= 1.2.2
+# uv is faster than pip for installing Python packages.
+uv ~= 0.5.30
+pytest-json-report
+
```
i-chaochen, line 20: I just notice jax/jaxlib also needs `pip install "numpy>=2.0" "ml_dtypes>=0.5.0" "scipy>=1.14" opt_einsum` as well https://github.com/jax-ml/jax/blob/2951d463560c53c926047938351ac742b244778c/setup.py#L65-L69 we need to include these into here I suppose? https://github.com/jax-ml/jax/blob/2951d463560c53c926047938351ac742b244778c/jaxlib/setup.py#L68-L70

i-chaochen (commented):
I just notice jax still neeeds `pip install "numpy>=2.0" "ml_dtypes>=0.5.0" "scipy>=1.14" opt_einsum` as well, is it already included

hsharsha:
@charleshofer  @mminutoli @JehandadKhan  Please let me know if we can merge this?
Is AWS CLI needed to upload jax whls to s3? Are we going to continue with that  release process? Also need to add device-all for multi arch support

.github/workflows/build_hermetic_ub22_docker.yml
```diff
@@ -0,0 +1,74 @@
+name: Build ROCm and TheRock JAX and XLA Docker images
+
+on:
+  push:
+    branches:
+      - hhs/add_docker_for_jax_xla
```
charleshofer: Looks like this needs to change so that we will run this workflow on pushes to the `rocm-jax-infra` branch

.github/workflows/build_hermetic_ub22_docker.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          persist-credentials: false
+      - name: Authenticate to GitHub Container Registry
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" \
+            | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+      - name: Build ROCm JAX docker image
+        working-directory: ./docker
+        run: |
+          IMAGE="ghcr.io/rocm/jax-ubu22.rocm${ROCM_VERSION}"
```
charleshofer: We already have a workflow that builds an image with this name. This workflow is going to conflict with that one. Also, we were using this name `jax-ubu22.rocmX.Y.Z` for images that already contained a complete installation of JAX intended to be used by customers/users. I very strongly recommend against using this name.

Also, why are we using Ubuntu 22 rather than Ubuntu 24?

.github/workflows/build_hermetic_ub22_docker.yml
```diff
@@ -0,0 +1,74 @@
...
+
+          docker build -f Dockerfile.rocm \
+            --build-arg ROCM_VERSION="${ROCM_VERSION}" \
+            -t "${IMAGE}:latest" .
+
+      - name: Push ROCm JAX docker image
+        working-directory: ./docker
+        run: |
+          IMAGE="ghcr.io/rocm/jax-ubu22.rocm${ROCM_VERSION}"
```
charleshofer: nit: We keep setting `IMAGE` again and again, can we stick it in the `env:` section?

docker/Dockerfile.theRock
charleshofer, line 1: Both Dockerfiles are missing a few things that we need to make CI work in upstream JAX:
1. The Github CLI (@mminutoli do we actually still need this one now that we're using AWS/S3?)
2. The AWS CLI
3. The Google Cloud CLI
4. Bazelisk

You can look at our old image on [how to install those](https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/Dockerfile.base-ubu24#L58), but if we can do the installs via `apt` I think that's preferrable.

docker/Dockerfile.theRock
```diff
@@ -0,0 +1,70 @@
+################################################################################
+ARG BASE_IMAGE=ubuntu:22.04@sha256:58b87898e82351c6cf9cf5b9f3c20257bb9e2dcf33af051e12ce532d7f94e3fe
+FROM $BASE_IMAGE AS devel
+# See https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact
+# on why we cannot reference BASE_IMAGE again unless we declare it again.
+################################################################################
+
+# TheRock nightly index URL (GPU-family-specific).
+#ARG THEROCK_INDEX_URL="https://repo.amd.com/rocm/whl-multi-arch/"
```
charleshofer: nit: Code commented out. Either use multi-arch and delete the 950 line, or use 950 and delete thisline.

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,53 @@
+################################################################################
+ARG BASE_IMAGE=ubuntu:22.04@sha256:58b87898e82351c6cf9cf5b9f3c20257bb9e2dcf33af051e12ce532d7f94e3fe
```
charleshofer: I'm not a fan of the whole `BASE_IMAGE` thing. It makes Dockerfiles really hard to manage IMO because it makes readers uncertain on what we're actually using to build our images. Can we just stick our preferred Ubuntu base image right in the `FROM ...` line and get rid of the whole `BASE_IMAGE` thing?

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,53 @@
...
+COPY builder.packages.txt /builder.packages.txt
+
+RUN /setup.sources.sh && /setup.packages.sh /builder.packages.txt
+
+# Setup Python
+COPY setup.python.sh /setup.python.sh
+COPY builder.requirements.txt /builder.requirements.txt
+
+# Since we are using python3.12 as the default python version, we need to
+# install python3.12 last for now.
```
charleshofer: If our intent is to make this image the image to end all images and use it everywhere, we should install all of the relevant Python versions for JAX, not just Python 3.12. See here https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/Dockerfile.base-ubu24#L26

docker/Dockerfile.rocm
mminutoli: Maybe ubuntu24.04

by pinning the sha we are not getting security updates.

docker/Dockerfile.theRock
```diff
@@ -0,0 +1,70 @@
+################################################################################
+ARG BASE_IMAGE=ubuntu:22.04@sha256:58b87898e82351c6cf9cf5b9f3c20257bb9e2dcf33af051e12ce532d7f94e3fe
```
mminutoli: Same comment as before.

docker/Dockerfile.theRock
```diff
@@ -0,0 +1,70 @@
...
+FROM $BASE_IMAGE AS devel
+# See https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact
+# on why we cannot reference BASE_IMAGE again unless we declare it again.
+################################################################################
+
+# TheRock nightly index URL (GPU-family-specific).
+#ARG THEROCK_INDEX_URL="https://repo.amd.com/rocm/whl-multi-arch/"
+ARG THEROCK_INDEX_URL="https://repo.amd.com/rocm/whl/gfx950-dcgpu/"
```
mminutoli, line 4: @gulsumgudukbay is this what they want us to use now?

charleshofer (changes requested):
Thanks for putting this together @hsharsha. I'm hoping that this can simplify our infra a bit. Just a few things,

1. IIUC, this is meant to be a replacement for all of our other Docker images. As part of this PR, can you remove all of the other Dockerfiles in the `/docker` directory as well as the `.github/workflows/build-base-docker.yml` workflow.
2. Can you fill in the description section of this PR and explain there why we need it?
3. @mminutoli How are we going to manage all the bandages that we're sticking on ROCm (MIOpen fix, ROCR fix, etc) to get the JAX tests to pass? I guess just add them to the Dockerfiles here?

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,53 @@
...
+# TODO(b/376338367): switch to pyenv.
+RUN /setup.python.sh python3.12 /builder.requirements.txt
```
mminutoli: We need at least 4 versions of python to run pytest: 3.11, 3.12, 3.13, 3.14

Upstream has also 3.13-nogil and 3.14-nogil

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,53 @@
...
+# TF does not support python3.13.
+RUN ln -sf /usr/bin/python3.12 /usr/bin/python3
+RUN ln -sf /usr/bin/python3.12 /usr/bin/python
+RUN ln -sf /usr/lib/python3.12 /usr/lib/tf_python
+
+# Setup ENV variables for tensorflow pip build
+ENV TF_NEED_ROCM=1
+ENV TF_ROCM_GCC=1
+ENV ROCM_TOOLKIT_PATH=${ROCM_PATH}
+
```
mminutoli, line 117: None of our fixes landed in 7.2.{1,2,3,4}... we need our custom MIOpen and the fix to the runtime that's in our docker.

docker/setup.rocm.sh
```diff
@@ -0,0 +1,152 @@
...
+            true
+    )
+    if [[ ${#MIOPENKERNELS[@]} -gt 0 ]]; then
+        DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated "${MIOPENKERNELS[@]}"
+    fi
+
+    #install hipblasLT if available
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated hipblaslt-dev || true
+
+elif [[ "$DISTRO" == "el7" ]]; then
```
mminutoli: why do we need to carry all of this baggage? do we need to test on el7 / el8?

docker/setup.sources.sh
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+# ==============================================================================
+#
+# Sets up custom apt sources for our TF images.
+
+# Prevent apt install tzinfo from asking our location (assumes UTC)
+export DEBIAN_FRONTEND=noninteractive
+
+# Set up shared custom sources
+apt-get update --allow-insecure-repositories
```
mminutoli: why allow-insecure? are we having certificate issues?

i-chaochen:
This aims to be the unified one for upstream JAx and XLA CI (at least in XLA upstream on JAx and XLA pipelines first), QA testing and dev. Not for the JAx releases.

mminutoli (changes requested):
We also need the google cloud tool to run our CI:

https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/Dockerfile.base-therock-ubu24#L82

and AWS tool https://github.com/ROCm/rocm-jax/blob/33b5a722cf84e1a0312156d869ec56764b2e2765/docker/Dockerfile.base-therock-ubu24#L100

docker/Dockerfile.theRock
mminutoli: We don't use gh cli in CI anymore. Not upstream at least. it's all through pip and the S3 bucket.

docker/Dockerfile.theRock
draganmladjenovic: The problem is that some tests run with default constructed debug_options. So we end up with in-process lld turned off for that tests. Once I remove that, this should not be needed.

draganmladjenovic: If we are doing symlinks here. Why not just use multiarch and patch symlinks that are missing?

docker/Dockerfile.rocm
alekstheod: Wouldn't it be easier to build using hermetic therock dependency throught:
--repo_env=ROCM_DISTRO_URL=url_to_the_rock_tarball? 
So we don't create a new image when switching over therock version?

docker/Dockerfile.theRock
```diff
@@ -0,0 +1,70 @@
...
+# TheRock version number (e.g. "7.13.0a20260401" or "" for latest).
+ARG THEROCK_VERSION="7.14.0"
+
+# Install devtoolset build dependencies
+COPY setup.sources.sh /setup.sources.sh
+COPY setup.packages.sh /setup.packages.sh
+COPY builder.packages.txt /builder.packages.txt
```
alekstheod: Can we get rid of these scripts? I think they just complicate the building. Instead lets just install the pkgs?

alekstheod:
Here is the clean-up pr: https://github.com/ROCm/rocm-jax/pull/483

.github/workflows/build_hermetic_ub22_docker.yml
alekstheod: Adjusted

.github/workflows/build_hermetic_ub22_docker.yml
alekstheod: Adjusted

.github/workflows/build_hermetic_ub22_docker.yml
alekstheod: I refactored the workflow to use the matrix. Reducing the number of repetitions

docker/Dockerfile.rocm
alekstheod: Adjusted

docker/Dockerfile.theRock
alekstheod: Done

docker/Dockerfile.theRock
alekstheod: This is used to not be asked for users input

charleshofer (changes requested):
This is way cleaner. Thanks for doing that. I believe that we still need to add the AWS and Google Cloud CLIs to the images so that JAX CI can read and write build artifacts to object storage.

alekstheod:
> AWS and Google Cloud CLIs

Added

charleshofer (approved):
Thanks for adding those. Looks good to me. You might want to wait for Marco's approval as well, though.

.github/workflows/build_hermetic_ub24_docker.yml
```diff
@@ -0,0 +1,79 @@
...
+      - 'docker/Dockerfile.rocm'
+      - 'docker/Dockerfile.theRock'
+      - 'docker/python_requirements.txt'
+      - '.github/workflows/build_hermetic_ub22_docker.yml'
+  pull_request:
+    paths:
+      - 'docker/Dockerfile.rocm'
+      - 'docker/Dockerfile.theRock'
+      - 'docker/python_requirements.txt'
+      - '.github/workflows/build_hermetic_ub22_docker.yml'
```
mminutoli, line 18: Should this be `ubu24` ?

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,115 @@
...
+# Prevent apt install tzinfo from asking our location (assumes UTC)
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Setup apt sources and install builder packages
+RUN apt-get update && \
+    apt-get install -y software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
```
mminutoli, line 11: is this stuff really needed?

for example, why build-essential? Doesn't the hermetic environment have all we need?

docker/Dockerfile.rocm
```diff
@@ -0,0 +1,115 @@
...
+        hipsparse-dbgsym \
+        hsa-rocr-dbgsym \
+        rccl-dbgsym \
+        rocfft-dbgsym \
+        rocm-core-dbgsym \
+        rocm-smi-lib-dbgsym \
+        rocprofiler-sdk-dbgsym \
+        rocsparse-dbgsym
+
+# Setup links for TensorFlow to compile.
```
mminutoli, line 105: This does not have our custom miopen nor our fixes to the [rocr runtime](https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/Dockerfile.base-ubu24#L182). I don't think they made it to the 7.2.4 release and we should still have failures without them.

.github/workflows/build_hermetic_ub24_docker.yml
alekstheod: Done

docker/Dockerfile.rocm
alekstheod: We still need some things, either for debugging or for supporting local sysroot local toolchain and so on.

docker/Dockerfile.rocm
alekstheod: I added miopen patch. The rocr_runtime seems to be already in.

mminutoli (approved):

docker/patches/build_miopen.sh
```diff
@@ -0,0 +1,74 @@
...
+    zlib1g-dev \
+    build-essential \
+    cmake \
+    git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Clone the repository with the fix
+git clone https://github.com/ROCm/rocm-libraries.git \
+    --branch fix/stack-overflow-kernel-init \
```
i-chaochen, line 37: For this miopen patch, why don't we use the patch atop release branch instead of using a customized branch?

alekstheod: I took it as it is from the docker files used before. I think this question shall be answered by the jax team @mminutoli @charleshofer

PR #462, opened 2026-06-12
Remove MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE from base image

magaonka-amd:
- Drop the MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE env var (and its comment) from docker/Dockerfile.base-ubu24
- This workaround forces MIOpen to skip the shipped PerfDb entry and run a live auto-tune search (then write back) for every convolution, making solver/kernel selection non-deterministic across runs and concurrent pytest-xdist workers
- That non-determinism is a likely source of flaky conv test failures (e.g. tests/batching_test.py::testConvGeneralDilated); removing it restores deterministic selection of the shipped find-db configs


psanal35 (approved):
LGTM.

PR #465, opened 2026-06-12
Disable the MIOpen user level database

mminutoli:
We suspect this is the root cause of flackyness.

We want to build containers and check with upstream CI.


magaonka-amd:
I see you have pushed docker , and CI cont looked okay for now. lets watch nightly too before approving this. but for now it is looking promising. thnaks.

PR #483, opened 2026-07-09
Clean-up docker files

alekstheod:
Simplify docker files


docker/Dockerfile.rocm
```diff
@@ -73,7 +73,20 @@ RUN ROCM_VERS=${ROCM_VERSION} && \
...
+        rocm-core-asan-dbgsym \
+        rocm-core-dbgsym \
+        roctracer-asan-dbgsym \
+        roctracer-dbgsym \
+        rocfft-asan-dbgsym \
+        rocfft-dbgsym \
+        rocrand-asan-dbgsym \
+        rocrand-dbgsym \
+        rocsparse-asan-dbgsym \
+        rocsparse-dbgsym
```
nurmukhametov: I am not sure this list is enough, I remember installing `hip-runtime-amd-dbgsym` for some stack traces.

alekstheod: How can we know which pkgs we would need? I will add this if we would need more lets append the list later?

nurmukhametov: I think that we need dbgsym packages for all *.so files from the deps we use in our CI. I believe these are missing at the moment:
```bash
          hipblas-dbgsym \
          hipfft-dbgsym \
          hipsparse-dbgsym \
          hipsolver-dbgsym \
          rccl-dbgsym \
          rocprofiler-sdk-dbgsym \
          rocm-smi-lib-dbgsym \
          hip-runtime-amd-dbgsym \
          comgr-dbgsym
```

Moreover, I don't think we need `-asan-dgbsym` packages at the moment. I believe they are needed only if we install and use asan versions of the rocm packages, e.g., `hip-runtime-amd-asan`.

hsharsha (approved):

nurmukhametov (approved):

PR #484, opened 2026-07-13
Override system python3 in manylinux containers

mminutoli:
## Summary
- Symlink a modern Python to `/usr/local/bin/python3` in both manylinux_2_28 container images (ROCm and TheRock)
- The manylinux_2_28 base image (AlmaLinux 8) ships Python 3.6 as `python3`, which lacks `typing.TypedDict` (added in 3.8)
- Third-party CI actions (e.g. `google-ml-infra/actions/ci_connection`) run `which python3` and fail with `ImportError: cannot import name 'TypedDict'`


magaonka-amd (approved):
nicee

PR #485, opened 2026-07-13
[release] Prep rocm-jaxlib-v0.11.0-alpha: TheRock 7.14.0rc3 images

Ruturaj4:
## Motivation
Make TheRock 7.14.0rc3 the default ROCm for the rocm-jaxlib-v0.11.0-alpha release images.

## Technical Details
- build-base-docker.yml: bump TheRock 7.13.0 -> 7.14.0rc3 and point the pinned entry at the prereleases index (rocm.prereleases.amd.com/whl-multi-arch/, where rc builds live). In-code TODO to revert the index to repo.amd.com/rocm/whl-multi-arch/ once 7.14 is GA.
- Consolidate the TheRock version+index into a single `therock-config` setup job that both the base and manylinux matrices consume via fromJSON, so an rc bump is one edit and the matrices can't drift.

## Test Plan / Result
Built the TheRock plugin/pjrt against 7.14.0rc3 locally and ran the full DLM suite on gfx950/MI355X: 13/13 pass. CI on this branch rebuilds/republishes the images.

## Submission Checklist
- [X] Look over the contributing guidelines.

### DIFF
--- a/.github/workflows/build-base-docker.yml
+++ b/.github/workflows/build-base-docker.yml
@@ -38,6 +38,31 @@ on:
       - 'docker/manylinux/*'
 
 jobs:
+  therock-config:
+    # Single source of truth for the TheRock ROCm version + pip index, consumed
+    # by both the base and manylinux TheRock jobs so they can't drift.
+    # TODO: when TheRock 7.14 is GA, bump the pinned entry to the final version
+    # and switch its index back to https://repo.amd.com/rocm/whl-multi-arch/
+    # (the prereleases index is for rc builds only).
+    if: >-
+      github.event_name != 'workflow_dispatch' ||
+      inputs.image-set == 'therock' ||
+      inputs.image-set == 'both'
+    runs-on: ubuntu-latest
+    outputs:
+      entries: ${{ steps.set.outputs.entries }}
+    steps:
+      - id: set
+        run: |
+          NIGHTLY="https://rocm.nightlies.amd.com/whl-multi-arch/"
+          PREREL="https://rocm.prereleases.amd.com/whl-multi-arch/"
+          {
+            echo 'entries<<JSON'
+            echo "[{\"therock-version\": \"\", \"therock-index-url\": \"$NIGHTLY\"},"
+            echo " {\"therock-version\": \"7.14.0rc3\", \"therock-index-url\": \"$PREREL\"}]"
+            echo 'JSON'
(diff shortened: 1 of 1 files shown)


docker/manylinux/Dockerfile.jax-manylinux_2_28-therock
```diff
@@ -20,8 +20,12 @@ RUN --mount=type=cache,target=/var/cache/dnf \
 # a /opt/python/ build and symlink its rocm-sdk onto PATH (not a /opt/rocm
 # symlink, just build plumbing so `rocm-sdk` is callable below).
 RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
+    case "${THEROCK_INDEX_URL}" in \
+        *.html) PIP_SRC="--find-links" ;; \
```
gulsumgudukbay: The *.html) PIP_SRC="--find-links" branch never runs; both index URLs end in /, so it always resolves to --index-url. If this is for the future, add a one-line comment; otherwise please remove it.

.github/workflows/build-base-docker.yml
```diff
@@ -158,13 +158,13 @@ jobs:
       fail-fast: false
       matrix:
         install-llvm: [true, false]
-        therock-version: ["", "7.13.0"]
+        therock-version: ["", "7.14.0rc3"]
```
gulsumgudukbay: This and line 259 is duplicates. it would be safer if we have a single source of truth and reuse that so that we don't forget to bump up one or the other. looks like we will do these every rc build/release of theRock.

docker/manylinux/Dockerfile.jax-manylinux_2_28-therock
gulsumgudukbay: also this is a brittle logic in my opinion. it would be nice if we coupld prefer passing the pip source flag explicitly from the workflow.

gulsumgudukbay (changes requested):
the PR body states 7.14.0rc1 -> 7.14.0rc3 with "prereleases index unchanged," but the diff actually goes from 7.13.0 to 7.14.0rc3 and switches the index URL from repo.amd.com/rocm/whl-multi-arch/ to rocm.prereleases.amd.com/whl-multi-arch/. 
Please update the  description. Also now that we are using prereleases index, we should add a TODO to revert it back to repo.amd once the release is done.

gulsumgudukbay (approved):

PR #496, opened 2026-07-27
Move TheRock ROCm install to the last layer in the therock docker images

magaonka-amd:
Pure layer reordering in the two TheRock docker images so the ROCm install is
the last non-metadata step.


tsrw2048 (approved):

magaonka-amd:
merging this to test rock changes in downstream. before we flip the switch in upstream.

PR #497, opened 2026-07-28
Add rocmless docker image

alekstheod:
Add rocmless docker image for CI infra


magaonka-amd (approved):
Thanks looks good to me.

PR #500, opened 2026-08-11
Add rocm_tag to pytest DB import dedup identity

psanal35:
We now publish two nightly ROCm tags. The dedup check in `find_existing_run_id` ignored `rocm_tag`, causing the second tag’s results to be dropped as duplicates.

This PR adds `rocm_tag` as a sixth identity field. The two nightly tags are stored as separate runs, while re-ingesting the same run and tag remains deduped.

No schema change is required: `rocm_tag` is already stored, and runs are ingested sequentially, so the code update is sufficient.


magaonka-amd (approved):
okay LGTM.

PR #502, opened 2026-08-17
[jax] Use prebuilt LLVM 18.1.8 for JAX 0.10.0

erman-gurses:
## Summary

Replace the LLVM 18.1.8 source build in the JAX 0.10.0 manylinux
Docker image with the official prebuilt LLVM 18.1.8 binary.

JAX 0.10.0 uses `clang_local` and expects Clang at:

`/usr/lib/llvm-18/bin/clang`

The prebuilt archive is therefore extracted to the same location, preserving
the existing compiler path while avoiding the full LLVM source build.

The GCC toolchain configuration previously supplied through
`GCC_INSTALL_PREFIX` is preserved with `clang.cfg` / `clang++.cfg`:

`--gcc-toolchain=/opt/rh/gcc-toolset-14/root/usr`

The LLVM archive is pinned by SHA256.

## Validation

Validated using the same `quay.io/pypa/manylinux_2_28_x86_64` base image.

Observed:

```text
/tmp/llvm.tar.xz: OK
clang version 18.1.8
InstalledDir: /usr/lib/llvm-18/bin
Configuration file: /usr/lib/llvm-18/bin/clang.cfg
```

This change is required for https://github.com/ROCm/TheRock/pull/7029.

### DIFF
--- a/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock
+++ b/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock
@@ -33,22 +33,27 @@ RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
 RUN ln -s /opt/python/cp312-cp312/bin/python3 /usr/local/bin/python && \
     python -m ensurepip && python -m pip install auditwheel
 
-# Install LLVM 18 from source (manylinux has no apt):
+# Install prebuilt LLVM 18.1.8 for clang_local.
 RUN --mount=type=cache,target=/var/cache/dnf \
-    dnf install -y wget && dnf clean all
-RUN mkdir /tmp/llvm-project && \
-    wget -qO - https://github.com/llvm/llvm-project/archive/refs/tags/llvmorg-18.1.8.tar.gz \
-        | tar -xz -C /tmp/llvm-project --strip-components 1 && \
-    mkdir /tmp/llvm-project/build && \
-    cd /tmp/llvm-project/build && \
-    cmake -DLLVM_ENABLE_PROJECTS='clang;lld' \
-          -DCMAKE_BUILD_TYPE=Release \
-          -DCMAKE_INSTALL_PREFIX=/usr/lib/llvm-18/ \
-          -DGCC_INSTALL_PREFIX=/opt/rh/gcc-toolset-14/root/usr \
-          ../llvm && \
-    make -j$(nproc) && \
-    make -j$(nproc) install && \
-    rm -rf /tmp/llvm-project
+    dnf install -y wget xz ncurses-compat-libs && \
+    dnf clean all
+
+RUN wget -q \
+      "https://github.com/llvm/llvm-project/releases/download/llvmorg-18.1.8/clang+llvm-18.1.8-x86_64-linux-gnu-ubuntu-18.04.tar.xz" \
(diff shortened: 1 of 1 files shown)


docker/manylinux/Dockerfile.jax-manylinux_2_28-therock
```diff
@@ -33,22 +33,27 @@ RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
...
+    tar -xJf /tmp/llvm.tar.xz \
+      -C /usr/lib/llvm-18 \
+      --strip-components=1 && \
+    rm /tmp/llvm.tar.xz && \
+    printf '%s\n' \
+      '--gcc-toolchain=/opt/rh/gcc-toolset-14/root/usr' \
+      > /usr/lib/llvm-18/bin/clang.cfg && \
+    cp /usr/lib/llvm-18/bin/clang.cfg \
+      /usr/lib/llvm-18/bin/clang++.cfg && \
+    /usr/lib/llvm-18/bin/clang --version
```
gulsumgudukbay, line 56: this exits 0 whether or not clang.cfg was loaded, and the entire gcc-toolchain pin rides on it. A clang bump or a config-dir change would silently drop the flag and you'd only find out via a confusing compile error later. 
```suggestion
    /usr/lib/llvm-18/bin/clang --version | grep -q "Configuration file:"
```

gulsumgudukbay (changes requested):
Overall I think the change is correct for 0.10.0 and 0.10.1 only. On 0.10.2 and 0.11.0, build:rocm_release_wheel --config=rocm_clang_hermetic, so those branches need the block deleted, not swapped, they never touch /usr/lib/llvm-18

I have one concern about this PR: The extracted tarball is considerably larger than the old make install tree since it ships all the static libs and tools. It's a throwaway CI image so probably irrelevant, but pruning to bin/clang*, bin/ld.lld and lib/clang/18 would keep it lean if disk on the builders ever bites.

This change is required for https://github.com/ROCm/TheRock/pull/7029. 
cc: @i-chaochen , @draganmladjenovic , @mminutoli  could you all please also review this PR?

erman-gurses:
> Overall I think the change is correct for 0.10.0 and 0.10.1 only. On 0.10.2 and 0.11.0, build:rocm_release_wheel --config=rocm_clang_hermetic, so those branches need the block deleted, not swapped, they never touch /usr/lib/llvm-18
> 
> I have one concern about this PR: The extracted tarball is considerably larger than the old make install tree since it ships all the static libs and tools. It's a throwaway CI image so probably irrelevant, but pruning to bin/clang*, bin/ld.lld and lib/clang/18 would keep it lean if disk on the builders ever bites.
> 
> This change is required for [ROCm/TheRock#7029](https://github.com/ROCm/TheRock/pull/7029). cc: @i-chaochen , @draganmladjenovic , @mminutoli could you all please also review this PR?

That is my plan -  I’ll keep the prebuilt LLVM 18.1.8 change limited to 0.10.0/0.10.1. For 0.10.2/0.11.0 I’ll remove the LLVM source-build block entirely since those release-wheel configs use rocm_clang_hermetic.

gulsumgudukbay (approved):

mminutoli:
Closing and reopening to retrigger CI.

